PR #41, opened 2017-04-21
mappers: Fixed companies update domain hook condition

mmadariaga:
It was being called only on company updates


mmadariaga:
@cruzccl Removed trailing whitespace

PR #43, opened 2017-04-21
mappers: Ensure valid domain value and remove unnecessary mapper logic

mmadariaga:
Added sip domain value verification in the model.
Removed empty sip domain value checker in the mapper because It's not necessary any more.

### DIFF
--- a/library/IvozProvider/Model/Companies.php
+++ b/library/IvozProvider/Model/Companies.php
@@ -21,6 +21,7 @@ namespace IvozProvider\Model;
 
 class Companies extends Raw\Companies
 {
+    const EMPTY_DOMAIN_EXCEPTION = 2001;
 
     /**
      * This method is called just after parent's constructor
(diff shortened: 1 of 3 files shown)


library/IvozProvider/Model/Companies.php
```diff
@@ -154,6 +154,24 @@ public function getMusicClass()
...
+     * @return \IvozProvider\Model\Raw\Companies
+     */
+    public function setDomainUsers($data)
+    {
+        if (is_string($data)) {
+            $data = trim($data);
+        }
+
+        if (empty($data)) {
+            throw new \Exception("Domain can't be empty");
```
Kaian: From total ignorance: can we make this text i18n? I've seen it's displayed in the portal.

mmadariaga: I'm not sure. Let me check it out

Issue #48, opened 2017-04-24
portals: Make From and To date required (pricingPlans-company rel)


cruzccl:
They are not required now and makes call simulator fail (and underlying logic is not prepared either).

PR #61, opened 2017-04-25
portals: PricingPlansRelCompanies required fields

mmadariaga:
#48 validFrom and validTo fields are required now in database and admin

Database migration delta will update any null value


mmadariaga:
@Kaian 
Default values have been modified in https://github.com/irontec/ivozprovider/commit/ad510313bde937348f079f2e1be5cd64ab104de2

Please check it out

PR #62, opened 2017-04-25
agi: error notifications over progress feature

Kaian:
This introduces a new Brand and Company feature 'Voice Notifications'. When
enabled, some errors will play a locution to the caller over progress (not
answering the call).

This provides extra information to the administrators in order to detect why
a call is not being processed.

It may also introduce some other weird scenarios where a user fordwards its
call to a destination that can not actually call and the locution is played to
the caller rather than the user who configured the call fordward (at least
provides extra information of why the fordward is not being done).

### DIFF
--- a/asterisk/agi/library/Agi/Action/ExternalFriendCallAction.php
+++ b/asterisk/agi/library/Agi/Action/ExternalFriendCallAction.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace Agi\Action;
+use \IvozProvider\Model\Features;
 
 /**
  * @class ExternalFriendCallAction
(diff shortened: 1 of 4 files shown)


Kaian:
There is [another place](https://github.com/irontec/ivozprovider/blob/bleeding/asterisk/agi/library/Agi/Action/ExternalDDICallAction.php#L41) in the sources where billable destination is checked but I'm not sure if we should notificate that.

When a **Company Administrator** configures an external number forward (for example on holidays filter) and enters a number it's company can't call (because there's no way to price that call), it will be directly rejected instead of playing the notification to the caller. I have skipped this on purpose because in this scenarios the caller is most probably external to the company and will not provide the extra information the notification provides.

mmadariaga:
@Kaian Out of interest. Is there any kind of notification/check when a non callable external number is setted?

Kaian:
**At PBX level**, there are two types of validations before placing an external call:
 - **ACL**: Based on User/Friend configured ACL, not applicable to Call Forwards configured by Company admins
 - **Billing**: Based on configured prices per destination at Brand level (not applicable if all brand's contract peerings use external rating)

**At Web Interface  level**, there isn't any kind of warning or validation of what you enter as destination number. Companies administrators and Users can configure any external number destination.

I will add documentation links for both ACLs and Pricing Plans but documentation is right now broken :-\

PR #63, opened 2017-04-25
portals: Fixed php execution error

mmadariaga:
Filter/RouteTypes.php: Call to a member function on null 
when there was no company being emulated


mmadariaga:
The filter is linked in klear/model/DDIs.yaml, so any screen with a routeType column makes use of it. Brand configuration > DDIs` for instance (without company emulation)

mmadariaga:
You're right, that was the first thing I thought. I've done this way so that more "edge case" scenarios are covered. Companies that do not exist any more, etc. Maybe I've gone a little bit too far :)

Issue #55, opened 2017-04-24
Make faxes email address configurable by brand


cruzccl:
Make them use the same source address as voicemail emails do (Brands/From).

PR #65, opened 2017-04-25
as: use brand notification address for received faxes

Kaian:
When a Fax is received and the send email option is enabled, the PBX will send
a notification with the fax converted to PDF attached.

The goal of this change is to use the email notification name and address
configurable at brand screen to customize email From header. We have also used
the template logics like we already do for voicemail notifications.

This also removes the extra mail notification that was being sent when an
outgoing fax failed to be sent as it's already displayed in the screen that is
used to send the fax. If this is not clear enough we can change the description
of the send email Fax field to 'send received fax to email'.

This PR aims to fix issue #55

### DIFF
--- a/asterisk/agi/library/Agi/Action/FaxCallAction.php
+++ b/asterisk/agi/library/Agi/Action/FaxCallAction.php
@@ -188,30 +188,71 @@ class FaxCallAction extends RouterAction
         // TODO Check return value for errors (use exec instead of shell_exec?)
         // Remove received tif file after conversion
         unlink($faxTIF);
+
+        // Store fax pages
+        $faxIn->setPages($this->agi->getVariable("FAXOPT(pages)"));
+
         // Check if this fax is associated with an email address
         if ($fax->getSendByEmail()) {
             $this->agi->verbose("Sending Fax [faxInOut%d] to %s", $faxIn->getId(), $fax->getEmail());
 
-            // Get faxdata values for mail message body and subject
-            $faxSrc = $faxIn->getSrc();
-            $faxDst = $faxIn->getDst();
+            // Get Fax Company
+            $company = $fax->getCompany();
+
+            // Get defaults mail settings
+            $config = \Zend_Controller_Front::getInstance()->getParam('bootstrap');
+            $mail = $config->getOption('mail');
+            $default_fromname = $mail['fromname'];
+            $default_fromuser = $mail['fromuser'];
+            $voicemail = $config->getOption('faxmail');
+            $template = $voicemail['template'];
 
             // Create attachment for PDF file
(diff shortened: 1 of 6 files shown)


asterisk/agi/library/Agi/Action/FaxCallAction.php
```diff
@@ -188,30 +188,67 @@ public function processFaxInStatus()
...
+            $pdfname = sprintf("fax-%s-%s.pdf", $faxIn->getSrc(), $faxIn->getDst());
             $attach = new \Zend_Mime_Part(file_get_contents($faxPDF));
             $attach->type = "application/pdf";
             $attach->disposition = \Zend_Mime::DISPOSITION_ATTACHMENT;
             $attach->encoding = \Zend_Mime::ENCODING_BASE64;
-            $attach->filename = "fax-$faxSrc-$faxDst.pdf";
+            $attach->filename = $pdfname;
+
+            $fromName = $company->getBrand()->getFromName();
+            if (empty($fromName)) $fromName = $default_fromname;
```
mmadariaga: I would add curly brackets to every if condition

asterisk/agi/templates/faxmail/default/en/body
```diff
@@ -0,0 +1,11 @@
+Greetings!
```
mmadariaga: You can use Zend_View for this if you prefer. 
http://stackoverflow.com/questions/1218191/how-can-i-make-email-template-in-zend-framework

Kaian: We could do a new issue to replace current template model to Zend_View and standarize them everwhere. We're also using two files (subject and body) which I don't like too much (maybe using the first line of the template for subject, rest for body?).

This basic _replace plain/text_ mode is used for:
 - [Voicemail notifications](https://github.com/irontec/ivozprovider/blob/bleeding/asterisk/agi/application/controllers/VoicemailController.php#L55)
 - [Recordings Disk Usage](https://github.com/irontec/ivozprovider/blob/bleeding/kamailio/recordings/application/controllers/RecordingsController.php#L353)
 - And now Incoming Faxes

I prefer to start with all of them in the same way and change them afterwards in the same change to avoid multiple implementations for the same thing.

What do you think?

asterisk/agi/library/Agi/Action/FaxCallAction.php
Kaian: Sure! This was copy-_pasta_ from voicemail notifications where it's also inline hehe

asterisk/agi/templates/faxmail/default/en/body
mmadariaga: Agree

PR #66, opened 2017-04-26
portals: get provisioning template runner back

mmadariaga:
Fixed path and minor improvements

### DIFF
--- a/portals/application/bin/template.php
+++ b/portals/application/bin/template.php
@@ -1,7 +1,7 @@
 #!/usr/bin/php
 <?php
 /**
- *  * Script Para importar las BBDD de labayru
+ *  Provisioning template wrapper
  */
 date_default_timezone_set('UTC');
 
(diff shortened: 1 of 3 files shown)


portals/application/bin/template.php
```diff
@@ -12,12 +12,11 @@
 // Define path to application directory
 $route = '';
 if ( APPLICATION_ENV != 'localdev'){
```
Kaian: I think this environment no longer exists. We only use development and production.

portals/application/bin/template.php
```diff
@@ -47,25 +46,25 @@
 
 $scriptOptions = $application->getOption('import');
 
-$miClase = new miclase($argv[1]);
+$miClase = new TemplateWrapper($argv[1]);
```
Kaian: We could also go a bit further and also name the variable $wrapper :innocent:

portals/application/controllers/KlearCustomRunCodeController.php
```diff
@@ -77,8 +78,7 @@ protected function _runCode($type, $inputMac, $error = false){
                 $this->view->terminalModel = $terminalModelsModel;
 
                 $var = base64_encode(serialize($this->view->getVars()));
-                $command = "/usr/bin/php ".$route . DIRECTORY_SEPARATOR .$filename." ".$var." 2>&1"; 
-
+                $command = "/usr/bin/php ".$route . DIRECTORY_SEPARATOR .$filename." ".$var." 2>&1";
```
Kaian: Not sure what changed in this line :thinking: 
The contatenation spaces are random, as it seems everywhere in this code.
But don't worry, we'll fix that in a future (never).

portals/application/bin/template.php
mmadariaga: Sure :)

PR #78, opened 2017-05-03
scheme: added name field in fixedCosts

mmadariaga:
Models, mappers and klear configuration regenerated

portals: 
- Applying required modifications in workers. 
- Added some minor UX improvements (margins, etc.)
- Updated templates

Requires workers to be restarted


mmadariaga:
@cruzccl templade added :)

PR #84, opened 2017-05-05
as/portals: Faxes logic rework 

Kaian:
I have splitted the changes of this PR in two commits, please see the commit description text.

The first commit reworks AGIs logics for faxes.
The second commit replaces the only existing AMI request for a new ARI request removing all intermediate pieces (Gearmand, Job and workers, and so on)


library/Asterisk/ARI/Connector.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        $this->_user = $config->ari["userName"];
+        $this->_pass = $config->ari["password"];
+        $this->_port = $config->ari["port"];
+    }
+
+    private function _createConnectionHandler()
+    {
+        $ch = curl_init();
```
mmadariaga: We should eventually change this for something testable. https://github.com/guzzle/guzzle maybe?

library/Asterisk/ARI/Connector.php
```diff
@@ -0,0 +1,109 @@
...
+    private function _post($ch, $url, $postdata)
+    {
+        curl_setopt($ch, CURLOPT_POSTFIELDS, json_encode($postdata));
+        curl_setopt($ch, CURLOPT_URL, $url);
+        return curl_exec($ch);
+    }
+
+    public function sendFaxfileRequest(\IvozProvider\Model\Raw\FaxesInOut $faxfile)
+    {
+        $applicationSerersMapper = new \IvozProvider\Mapper\Sql\ApplicationServers();
```
mmadariaga: You've got a typo here applicationSer**v**ersMapper

library/Asterisk/ARI/Connector.php
```diff
@@ -0,0 +1,109 @@
...
+                    )
+            );
+
+            // Get Application server restAPI URL
+            $url = "http://" . $asAddress . ":" . $this->_port . "/ari/channels";
+            $this->_logger->log("Sending ARI Request to $url...", \Zend_Log::INFO);
+
+            // Reques a new call
+            $response = $this->_post($ch, $url, $post);
+            if ($response === false) {
```
mmadariaga: Could you check a truthy condition first in order to make else condition unnecessary?

library/Asterisk/ARI/Connector.php
```diff
@@ -0,0 +1,109 @@
...
+                $this->_logger->log("Error sending fax to $url:" . curl_error($ch), \Zend_Log::ERR);
+            } else {
+                $this->_logger->log("Request successfully enqueued.", \Zend_Log::INFO);
```
mmadariaga: Should we return status?

library/IvozProvider/Mapper/Sql/FaxesInOut.php
```diff
@@ -31,9 +31,8 @@ protected function _save(\IvozProvider\Model\Raw\FaxesInOut $model,
         $statusHaschanged = $model->hasChange("status");
 
         if ($isOutgoingFax && $statusHaschanged && $isPending) {
-            $faxSenderJob = new \IvozProvider\Gearmand\Jobs\FaxSender();
-            $faxSenderJob->setFaxInOut($model);
-            $faxSenderJob->send();
+            $ari = new \Asterisk\ARI\Connector();
+            $ari->sendFaxfileRequest($model);
```
mmadariaga: Given that this is not async anymore, we could throw an error here when something goes wrong

library/Asterisk/ARI/Connector.php
Kaian: If I disable this option to make curl_exec return boolean:
```
// curl_setopt($ch, CURLOPT_RETURNTRANSFER, TRUE);
```
Maybe something like this?

```
    $response = $this->_post($ch, $url, $post);
    if ($response === true) {
        $this->_logger->log("Request successfully enqueued to $url.", \Zend_Log::INFO);
        break;
    }
    $this->_logger->log("Error sending fax to $url:" . curl_error($ch), \Zend_Log::ERR);
} /* foreach loop */

return $response;

```

library/IvozProvider/Mapper/Sql/FaxesInOut.php
Kaian: You mean make Connector throwing Exceptions or just checking the boolean return that sendFaxfileRequest will return and throw an exception here with an error code (that can be translated in klear)?

library/Asterisk/ARI/Connector.php
Kaian: See below :arrow_down:

library/Asterisk/ARI/Connector.php
Kaian: :sweat_smile:

library/Asterisk/ARI/Connector.php
Kaian: Umm, seems ok to me. I think this is the first time a HTTP request is done from a mapper so we can change now or later, as you prefer.

mmadariaga: We can wait until composer dependencies are unified

library/Asterisk/ARI/Connector.php
mmadariaga: :+1:

Kaian:
Note for myself: You could at least use the same tag for the commit instead of faxes: and fax: :stuck_out_tongue:

library/IvozProvider/Mapper/Sql/FaxesInOut.php
mmadariaga: I would prefer to throw the exception within the Connector. You can register the error in portals/application/configs/clear/errors. Yaml (Requires the page to be reloaded, currently error translations are done with javascript)

Kaian: The i18n for the mappers errors is something I don't care too much. I see them more related with the portal than mappers.

The code in the other hand is what someone using the API (that will end in the same mapper) will take into account to check what has gone wrong.

Any suggestion for the error code number?

Kaian:
I have updated with some of the requested changes on Connector.php and also some header comments for the functions.

library/IvozProvider/Mapper/Sql/FaxesInOut.php
Kaian: If request fails in all Application Servers, connector will throw an exception. It seems that exception is not required to be handled in the mapper, so I skipped the try..catch blocks in this section.

See 7e79e21

Kaian:
Updated with your suggestions and a couple of minor fixes.

Thanks for your time Mikel :))

PR #96, opened 2017-05-17
portals: cache file thumbnails

### LINKED ISSUE #81: portals: store generated thumbnails
Some portal lists (p.e. Faxes or Invoices) have thumbnails for generated or uploaded PDFs that seems to be created every time the list is refreshed.

This also increases loading times a lot, as it seems edit/new screens of those lists are not loaded until all visible thumbnails has been rendered (you can try creating a new faxfile while the list contains ten or more already sent faxes).

The most traditional approach will be to generate thumbmails once and reuse them, but they must be linked somehow with the FSO because all related objects can be updated or deleted.

mmadariaga:
This PR aims to fix #81

**Requires klear to be updated.**

Faxes with no error are not editable anymore, read only screen is shown instead.
The pdf in faxes is not editable anymore
Fixed: set fax status to error when an exception is thrown
Removed invoice thumbnails from lists

---
This change is [Reviewable](https://reviewable.io/reviews/irontec/ivozprovider/96)

### DIFF
--- a/portals/application/configs/klear/FaxesList.yaml
+++ b/portals/application/configs/klear/FaxesList.yaml
@@ -115,6 +115,7 @@ production:
         options:
           title: _("Options")
           screens:
+            faxesOutView_screen: true
             faxesOutEdit_screen: true
           dialogs:
             forwardFax_dialog: true
@@ -137,6 +138,15 @@ production:
       class: ui-silk-eye
       disableSave: true
 
+    faxesOutView_screen:
+      <<: *faxesInOutEdit_screenLink
+      parentOptionCustomizer:
+        - IvozProvider_Klear_Options_OptionsCustomizerFilterForFax
+      title: _("View Outgoing %s %s", ngettext('Faxfile', 'Faxefiles', 0), "[format| (%item%)]")
+      filterField: faxId
+      class: ui-silk-eye
+      disableSave: true
+
     faxesOutNew_screen:
       <<: *faxesInOutNew_screenLink
       filterField: faxId
@@ -153,6 +163,8 @@ production:
 
     faxesOutEdit_screen:
       <<: *faxesInOutEdit_screenLink
+      parentOptionCustomizer:
+        - IvozProvider_Klear_Options_OptionsCustomizerFilterForFax
       forcedValues:
         type: 'Out'
       filterField: faxId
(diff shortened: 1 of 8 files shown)


portals/application/configs/klear/model/Invoices.yaml
```diff
@@ -92,17 +92,17 @@ production:
...
+#            target: invoicesPdfFilePreview_command
+#            type: command
+#            class: filePreview
+#            listController: 1
+#            external: 1
+#            props:
+#              width: 30
+#              height: 30
+#              crop: false
+#            onNull: hide
```
Kaian: Commented source code is not allowed. Remove it! <3

portals/application/configs/klear/FaxesList.yaml
```diff
@@ -137,6 +138,15 @@ production:
       class: ui-silk-eye
       disableSave: true
 
+    faxesOutView_screen:
+      <<: *faxesInOutEdit_screenLink
+      parentOptionCustomizer:
+        - IvozProvider_Klear_Options_OptionsCustomizerFilterForFax
+      title: _("View Incoming %s %s", ngettext('Faxfile', 'Faxefiles', 0), "[format| (%item%)]")
```
Kaian: I miss a translation of this string, or it was already translated?

If you rebase post 75a563f76de6aa46e7dc6ca213eac56f859ad8e8 translation files should be clean! :)

portals/application/configs/klear/model/Invoices.yaml
mmadariaga: Fair enough :)

portals/application/configs/klear/FaxesList.yaml
mmadariaga: Um, this is wrong actually. It should be "outgoing". I'll add the translation as you said

portals/application/languages/en_US/en_US.po
```diff
@@ -2030,7 +2035,7 @@ msgstr "Strategy"
 
 #, fuzzy
 msgid "ringall"
-msgstr "A todos"
+msgstr "Ring all"
```
Kaian: Extra translation, extra thanks!

Kaian:
Squash and :shipit:  !

PR #104, opened 2017-05-22
portals: Avoided main operator to emulate brand automatically

mmadariaga:
The brand linked to current god BrandURL was being automatically emulated

---
This change is [Reviewable](https://reviewable.io/reviews/irontec/ivozprovider/104)

### DIFF
--- a/library/IvozProvider/Klear/Auth/Adapter.php
+++ b/library/IvozProvider/Klear/Auth/Adapter.php
@@ -22,22 +22,23 @@ class Adapter implements \Klear_Auth_Adapter_KlearAuthInterface
 
     public function __construct(\Zend_Controller_Request_Abstract $request, \Klear_Model_ConfigParser $authConfig = null)
     {
-
-        if (!$authConfig->getProperty("brandId")) {
-            throw new \Klear_Exception_Default('brandId not on authConfig');
-        }
-
         $this->_username = $request->getPost('username', '');
         $this->_password = $request->getPost('password', '');
-        $brandMapper = new BrandsMapper();
-        $this->_brand = $brandMapper->find($authConfig->getProperty("brandId"));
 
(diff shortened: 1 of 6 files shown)


library/IvozProvider/Klear/Auth/Adapter.php
```diff
@@ -22,22 +22,18 @@ class Adapter implements \Klear_Auth_Adapter_KlearAuthInterface
...
-
-        if (!$authConfig->getProperty("brandId")) {
-            throw new \Klear_Exception_Default('brandId not on authConfig');
-        }
-
         $this->_username = $request->getPost('username', '');
         $this->_password = $request->getPost('password', '');
         $brandMapper = new BrandsMapper();
-        $this->_brand = $brandMapper->find($authConfig->getProperty("brandId"));
+        if ($authConfig->getProperty("brandId")) {
```
Kaian: `$brandMapper` declaration will make more sense inside the `if`condition ?

Default `$authConfig` parameter value (_null_) makes `if` invokation crash :open_mouth:

portals/application/controllers/KlearCustomExtraAuthController.php
```diff
@@ -203,9 +201,9 @@ protected function _getData()
         $result = $storage->write($this->_user);
 
         return array(
-                'result'=>$result,
-                'newIden' => $remoteId
-                );
+            'result'=>$result,
+            'newIden' => $remoteId
+        );
```
Kaian: spaces and arrows :innocent:

library/IvozProvider/Klear/Auth/Adapter.php
mmadariaga: It seems to be an impossible scenario, but the default value is forced at interface level. I'll add a security check just in case

PR #105, opened 2017-05-22
friend: allow outgoing external calls with any of company's DDI

### LINKED ISSUE #86: friends: Allow different outgoing DDIs
Friends were designed in order to allow:

- Making calls from IvozProvider to users of an external PBX.

- Making calls from that external PBX to IvozProvider users.

But we also provided another implicit feature: **allow external PBX users to make external calls using IvozProvider**. That's why they have:

- Outgoing DDI.
- Call ACLs.

As behind an external PBX they might be multiple users, it would be nice to allow them call with different outgoing DDIs.

Proposal:

- Convert Outgoing DDI to "Default Outgoing DDI".
- Add a new selector: "Respect presented CLID" enable/disable.

If the presented CLID matches one of the Companies DDI, IvozProvider will respect this CLID. If not, IvozProvider will use the Default Outgoing DDI.

**Note:** adding DDIs to a Company in order to use them as a valid Outgoing DDI (for a friend or a user) will lead to malfunction. **IvozProvider DDIs MUST enter IvozProvider**, there no "Outgoing DDIs" section.

Kaian:
If friend's presented DDI in PAI header matches one of the company's DDIs, that DDI will be used to place the external call. If it doesn't match, the fallback DDI will be used (that can be the default Company's DDI).

This PR aims to fix #86


asterisk/agi/library/Agi/Action/ExternalFriendCallAction.php
```diff
@@ -70,8 +70,27 @@ public function process()
             return;
         }
 
-        // Outgoing presentation
-        $ddi = $friend->getOutgoingDDI();
+        // Allow identification from any company DDI
+        $callerIdNum = $this->agi->getCallerIdNum();
+        $companyDDIs = $friend->getCompany()->getDDIs();
+        foreach ($companyDDIs as $companyDDI) {
+            // FIXME This does not support E.164 DDIs
```
cruzccl: I think that PAI/RPID number format should follow Friend's Country format. That way, using preferredToE164 function the result can be matched against E.164 DDI.

e.g. Friend country is Spain ---> PAI/RPID should follow Spain's calling format to be detected as a valid source.

Kaian: Updated! :clown_face:

PR #107, opened 2017-05-22
as: AoR extract function and URI params

### LINKED ISSUE #106: as: AoR extract function and URI params
_extractURI_ function in _asterisk/agi/library/Agi/Wrapper.php_ is responsible for extracting the AoR address of a transferrer from Referred-by header. This is not the most elegant way to achieve this, but until we find a better way, it does the job.

e.g.

Referred-By: <sip:myuac@mydomain.org> --> extracts myuac@mydomain.org

When the URI contains a param, this function fails and blind-xfers fail (e.g. Referred-By: <sip:3402@sip.genesis.u50.es;user=phone>).

Fix the regexp to parse URI params fine.

cruzccl:
This PR aims to fix #106.

---
This change is [Reviewable](https://reviewable.io/reviews/irontec/ivozprovider/107)

### DIFF
--- a/asterisk/agi/library/Agi/Wrapper.php
+++ b/asterisk/agi/library/Agi/Wrapper.php
@@ -304,7 +304,7 @@ class Agi_Wrapper
 
     public function extractURI($uri, $variable)
     {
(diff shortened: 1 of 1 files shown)


Kaian:
LGTM :shipit: :+1:

PR #108, opened 2017-05-22
portals: reviewed operator required fields

mmadariaga:
name and surname are not required anymore.
unified form styles
print username instead of name + surname in klear
advanced settings in operators 
fixed fatal error in PHP7


mmadariaga:
@cruzccl You got it

PR #109, opened 2017-05-25
mappers: removed username from users

mmadariaga:
Applied required changes in userweb and sql mapper


Kaian:
I was trying to review this one because I understand that it's database schema changes must be applied before #110 but the delta 047 here seems empty. 

Is that ok or am I missing something?

Thanks!

mmadariaga:
@Kaian 
Yes, generated models and mapper in #110 already have schema changes of this PR applied

library/IvozProvider/Mapper/Sql/Users.php
```diff
@@ -20,45 +20,52 @@
...
+        }
 
         if ($isNew) {
             // Sane defaults for hidden fields
-            if (!$model->hasChange('timezoneId'))
-                $model->setTimezoneId($model->getCompany()->getBrand()->getDefaultTimezoneId());
-
-            if (!$model->hasChange('exceptionBoosAssistantRegExp'))
-                $model->setExceptionBoosAssistantRegExp('');
+            $model->setTimezoneId(
```
cruzccl: All the _hasChange fiesta_ is to avoid overriding values in API created items.

I'm afraid they are still necessary, aren't they?

mmadariaga:
@cruzccl It should be already fixed, please check it out :)

PR #113, opened 2017-05-30
portals: Changes in numeric transformation generator

cruzccl:
Do not ask for national length as generated rules don't use it anymore.


portals/application/configs/klear/model/PeeringContracts.yaml
```diff
@@ -26,11 +26,11 @@ production:
             template: '%name%'
           order: name
     transformationRulesetGroupsTrunksId:
-      title: _('Transformation ruleset groups trunks')
+      title: _('Numeric transformation')
       type: select
       source:
         data: mapper
-        'null': _("Unassigned")
+        'null': _("E.164")
```
Kaian: If you're not willing to translate this string, dont use `_( )` here.

portals/application/configs/klear/model/PeeringContracts.yaml
```diff
@@ -26,11 +26,11 @@ production:
             template: '%name%'
           order: name
     transformationRulesetGroupsTrunksId:
-      title: _('Transformation ruleset groups trunks')
+      title: _('Numeric transformation')
```
Kaian: Is this already translated?

portals/application/configs/klear/model/TransformationRulesetGroupsTrunks.yaml
```diff
@@ -84,6 +84,7 @@ production:
               - calling_code
             template: '%name% (+%calling_code%)'
           order: name_${lang}
+      defaultValue: 70
```
Kaian: I don't like hardcoding values. For DDI's we use default company's countryId taken from auth (check DDIList.yaml), but Brands have no countryId, so I won't argue for this.

portals/application/configs/klear/model/PeeringContracts.yaml
cruzccl: Yes it is.

portals/application/configs/klear/model/PeeringContracts.yaml
cruzccl: You are right :)

PR #115, opened 2017-05-30
portals: PeerServer name is not needed anymore

cruzccl:
It was a required value but was not used anywhere. Now it is dinamically
generated so that its sister-table (LcrGateways) has an unambiguous name.


portals/application/modules/rest/controllersRaw/PeerServersController.php
```diff
@@ -586,16 +562,6 @@ public function optionsAction()
                     'required' => false,
                     'comment' => '',
                 ),
-                'name' => array(
```
mmadariaga: You need to apply this changes into rest/controllers/PeerServersController.php as well

mmadariaga:
We've some rest controllers are outdated:
- ApplicationServersController.php
- BrandOperatorsController.php
- CompanyAdminsController.php
- FixedCostsController.php
- MainOperatorsController.php
- PricingPlansRelCompaniesController.php
 
We'll need to update them and regenerate apidocs before 1.3 is released.
We can do it in a separate PR if you prefer

PR #126, opened 2017-06-05
portals: increased delete security for brands

mmadariaga:
Applied for companies as well

Fixed error when multiple companies were selected

### DIFF
--- a/library/IvozProvider/Klear/Filter/OutgoingDDI.php
+++ b/library/IvozProvider/Klear/Filter/OutgoingDDI.php
@@ -10,8 +10,13 @@ class IvozProvider_Klear_Filter_OutgoingDDI implements KlearMatrix_Model_Field_S
         $pk = $routeDispatcher->getParam("pk", false);
 
         // Only display DDIs belonging to edited company
-        $this->_condition[] = "`companyId` = '" . $pk . "'";
+        if (is_array($pk)) {
+            //Avoid multiple choice error on klear
+            $this->_condition[] = "1=2";
+            return true;
+        }
 
+        $this->_condition[] = "`companyId` = '" . $pk . "'";
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


library/IvozProvider/Klear/Filter/OutgoingDDI.php
```diff
@@ -10,8 +10,13 @@ public function setRouteDispatcher(KlearMatrix_Model_RouteDispatcher $routeDispa
         $pk = $routeDispatcher->getParam("pk", false);
 
         // Only display DDIs belonging to edited company
-        $this->_condition[] = "`companyId` = '" . $pk . "'";
+        if (is_array($pk)) {
+            //Avoid multiple choice error on klear
+            $this->_condition[] = "1=2";
```
Kaian: If there's no way to avoid the query, what about...
```
    $this->_condition[] = "FALSE";
```

mmadariaga: Umm, I like the fact that you can see that it's a condition itself easily

Kaian: It shouldn't. FALSE is a SQL keyword afaik

cruzccl: I don't understand the relationship between this feature and OutgoingDDI.php file :thinking:

mmadariaga: @Kaian I know, I just prefer it that way. It's kind of more universal. @cruzccl tiebreak this please :)

mmadariaga: @cruzccl Try to select a couple of companies and press delete button then

PR #127, opened 2017-06-06
portals: removed importers and exporters

mmadariaga:
Removed everywhere but Billable calls

Added some minor improvements into ImportFileController.php:
- Show the same field name as in exported csv
- Sort field lists

### DIFF
--- a/portals/application/configs/klear/PricingPlansRelTargetPatternsList.yaml
+++ b/portals/application/configs/klear/PricingPlansRelTargetPatternsList.yaml
@@ -39,7 +39,7 @@ production:
           pricingPlansRelTargetPatternsDel_dialog: true
           importPricingPlansRelPatterns_dialog: true
       csv:
-        active: true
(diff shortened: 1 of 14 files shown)


portals/application/configs/klear/PricingPlansRelTargetPatternsList.yaml
```diff
@@ -37,9 +37,9 @@ production:
           pricingPlansRelTargetPatternsNew_screen: true
         dialogs: 
           pricingPlansRelTargetPatternsDel_dialog: true
-          importPricingPlansRelPatterns_dialog: true
+          importPricingPlansRelPatterns_dialog: false
```
cruzccl: This one should be keeped, this is the only one that is really useful ;)

mmadariaga: Thanks. You got it

PR #130, opened 2017-06-06
portals: fixed add main operator regression

mmadariaga:
Fixed regression introduced in #108 
Main operators had no active value (null)


mmadariaga:
@cruzccl I thought that the alter table you suggested would end up in an error becouse the models/klear would try to insert null values. It's not the case, it turns out it does the trick, thanks :)

Issue #128, opened 2017-06-06
invoices: Improve logic to make them faster


cruzccl:
Generating a PDF invoice has 2 steps:

- Getting the calls that will be included.

- Generating the PDF using phppdf library and the invoices templates.

Both of the steps are too slow now.

The first step shouldn't be slow at all (it is a just a sql query), analyze why.

PR #132, opened 2017-06-07
workers: Improved invoice db related performance

mmadariaga:
The performance has been almost duplicated. DB accesses only represent 10% of the execution time nonetheless

This PR aims to close https://github.com/irontec/ivozprovider/issues/128

### DIFF
--- a/library/IvozProvider/Gearmand/Invoices/Generator.php
+++ b/library/IvozProvider/Gearmand/Invoices/Generator.php
@@ -3,12 +3,13 @@ namespace IvozProvider\Gearmand\Invoices;
 
 class Generator
 {
-
     protected $_invoiceId = null;
     protected $_logger = null;
     protected $_fixedCostTotal = 0;
     protected $_fixedCosts = array();
     protected $_totals = array();
+
+    protected $pricingPlanCache = [];
 
     public function __construct($invoiceId = null, $logger = null)
     {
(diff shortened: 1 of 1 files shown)


mmadariaga:
I've encreased call batch from 50 to 100, test it under PHP 5.6 please (Max memory issues?)

library/IvozProvider/Gearmand/Invoices/Generator.php
```diff
@@ -163,7 +165,14 @@ protected function _getCallData(\IvozProvider\Model\Invoices $invoice)
             );
             $this->_fixedCostTotal  += number_format(ceil($subTotal*10000)/10000, 4);
         }
+
+        $dbAdapter = $invoice->getMapper()->getDbTable()->getAdapter();
+        $updateCallsInvoiceIdQuery = 'UPDATE `kam_acc_cdrs` SET `invoiceId` = ' . $invoice->getId();
+        $updateCallsInvoiceIdQuery .= ' WHERE id in (%s)';
```
Kaian: Is there any difference between using the call ids array or the conditions already existing in `$where` variable?

The performance for these is minimal but I'm imaging myself debugging a query with a list of 5000 ids separeted by comas hehe.

library/IvozProvider/Gearmand/Invoices/Generator.php
```diff
@@ -174,12 +183,15 @@ protected function _getCallData(\IvozProvider\Model\Invoices $invoice)
 
             $offset += $limit;
 
+
             foreach ($calls as $call) {
                 if (!$call) {
+                    //WTF
                     $call = new \IvozProvider\Model\KamAccCdrs();
```
mmadariaga: Agree

library/IvozProvider/Gearmand/Invoices/Generator.php
mmadariaga: It's executed every loop, 100 calls actually. But sure, I can change it so that it's executed once everything is computed

Issue #15, opened 2017-02-27
Add support for generic Voicemails


Kaian:
Rigth now we have once voicemail per user (if it's enabled for that user).

This enhancement add support for generic voicemails where multiple terminals can subscribe to their status and/or check it.

This require:
 1) Add support for checking any voicemail (not just ours)
 2) Add support for MWI subscriptions

PR #134, opened 2017-06-09
agi: support extension in voicemail check services

### LINKED ISSUE #136: Listen to voicemails from any terminal
As mentioned in #15 , it would be nice to add the possibility to listen to your voicemail from any terminal, not only from yours.

The implementation would be like this:

- *93 (or any configured code) would access to your voicemail, without password.

- *93101 (*+<configured_code>+extension) would access to the voicemail of the user with extension 101, asking his password first.

This introduces passwords in voicemail access:

- Default password for every users: '1234'.

- In voicemail menu you can change this password, even delete it (leaving it blank).

- Password will only be asked when accessing with *93101, accessing to your voicemail with *93 will be password-free as it is now.

Kaian:
Prior to this change the Voicemail check service only allowed checking the caller voicemail.

This changes allows to provide an extra extension after the service code, to check other user's.
voicemail (p.e. *981001 to check user 1001 voicemail).

Users can still secure their voicemails using the voicemail settings menu and configure a password.
Password won't be requested when the caller checks its own voicemail.

This PR is related with #15 and fixes #136

### DIFF
--- a/asterisk/agi/library/Agi/Action/ServiceAction.php
+++ b/asterisk/agi/library/Agi/Action/ServiceAction.php
@@ -44,9 +44,42 @@ class ServiceAction extends RouterAction
     {
         // Local variables to improve readability
         $caller = $this->_caller;
+        $service = $this->_service;
+        $company = $caller->getCompany();
 
-        // Checkvoicemail for this user
-        $this->agi->checkVoicemail($caller->getVoiceMail());
+        /**
+         * Extract optional Voicemail Extension from dialed number
+         *
+         *               ServiceCode (up to 3 digits)
+         *                   ┌┴┐
+         *   $dialedExten = *CCCXXXXXXXX
+         *                      └───┬──┘
+         *                      VoicemailExtension (optional)
+         */
+        $dialedExten = $this->agi->getExtension();
+        $serviceCodeLen = strlen($service->getCode());
+        $vmExtension = substr($dialedExten, $serviceCodeLen + 1);
+
+        if (!empty($vmExtension)) {
+            $extension = $company->getExtension($vmExtension);
+
+            if (empty($extension)) {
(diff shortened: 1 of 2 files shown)


asterisk/agi/library/Agi/Action/ServiceAction.php
```diff
@@ -44,9 +44,27 @@ protected function _processVoiceMail()
     {
         // Local variables to improve readability
         $caller = $this->_caller;
+        $service = $this->_service;
+        $company = $caller->getCompany();
 
-        // Checkvoicemail for this user
-        $this->agi->checkVoicemail($caller->getVoiceMail());
+        // Check if the service has target extension
+        $exten = substr($this->agi->getExtension(), strlen($service->getCode()) + 1);
```
mmadariaga: This requires a comment, at least. It doesn't make any sense to me

Kaian: What about:

 ```
/**
 *             ServiceCode (up to 3 digits)
 *                 ┌┴┐
 * $dialedExten = *CCCXXXXXXXX
 *                    └───┬──┘
 *                    VoicemailExtension (optional)
 **/
$dialedExten = $this->agi->getExtension();
$serviceCodeLen = strlen(strlen($service->getCode());
$vmExtension = substr($dialedExten, $serviceCodeLen + 1);
```

cruzccl:
Implements #136

asterisk/agi/library/Agi/Action/ServiceAction.php
mmadariaga: Wow, more than expected :-)

Kaian:
Updated with the suggestions.

I have also added a couple of spaces to improve readability and some comments

PR #143, opened 2017-06-13
Fix notify uri

cruzccl:
This PR aims to fix the problem described in the first commit message.

### DIFF
--- a/scheme/deltas/057-fix-notify-uri.sql
+++ b/scheme/deltas/057-fix-notify-uri.sql
@@ -1,0 +1,7 @@
+DROP VIEW IF EXISTS `kam_users_dbaliases`;
+
+DELETE FROM `kam_version` WHERE table_name='kam_users_dbaliases';
+
+DROP VIEW IF EXISTS `kam_users_exten`;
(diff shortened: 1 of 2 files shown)


scheme/deltas/057-fix-notify-uri.sql
```diff
@@ -0,0 +1,7 @@
+DROP VIEW IF EXISTS `kam_users_dbaliases`;
+
+DELETE FROM `kam_version` WHERE table_name='kam_users_dbaliases';
+
+DROP VIEW IF EXISTS `kam_users_exten`;
+CREATE VIEW `kam_users_exten` AS SELECT T.name AS name, T.domain AS domain, E.number AS extension FROM Users U INNER JOIN Terminals T ON U.terminalId = T.id JOIN Extensions E ON E.id = U.extensionId;
```
Kaian: I will explicitly indicate INNER JOIN here. I also tent to sort ON clause first with the join table.

```
CREATE VIEW `kam_users_exten` AS SELECT T.name AS name, T.domain AS domain, E.number AS extension FROM Users U INNER JOIN Terminals T ON T.id = U.terminalId INNER JOIN Extensions E ON E.id = U.extensionId;
```

PR #140, opened 2017-06-12
mappers: changed pdf generator lib

mmadariaga:
Added HTML templates
Added klear help texts
Unified composer files
composer.lock is not ignored anymore
Updated documentation

This PR fixes https://github.com/irontec/ivozprovider/issues/138

PR #145, opened 2017-06-13
Simplify kam users view

cruzccl:
$terminalName -> $extension conversion is now made in GENERATE_PUBLISH route, so both main query and kam_users view can be simplified.


Kaian:
This MUST be merged after #140 due to scheme delta numbering.

PR #146, opened 2017-06-13
tests: add initial bbs SIP tests

Kaian:
Files for simple SIP tests using BBS tool.

This tests are extremely superficial and they should be couple with
traditional [SIPP](https://github.com/SIPp/sipp) tests.

### DIFF
--- a/tests/bbs/0500-test-extension.yaml
+++ b/tests/bbs/0500-test-extension.yaml
@@ -1,0 +1,28 @@
+# test-ddi.yaml
+#
+# Simple scenario for testing a call between alice and bob
+# Bob must have the extension 1001 assigned
+#
+# ----------------------------------------------------------------------------
+scenarios:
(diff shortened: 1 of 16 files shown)


tests/bbs/0500-test-extension.yaml
```diff
@@ -0,0 +1,28 @@
...
+          - wait: 2
+          - hangup
+          - waitfor: DISCONNCTD
+
+      - call_uas_bob:
+          - register:
+              <<: *bob_cred
+          - waitfor: INCOMING
+          - answer
+          - waitfor: DISCONNCTD
```
mmadariaga: Is this some SIP value or something? Otherwise we can use DISCONNECTED instead

Kaian: Yes, its generated by pjsip

All events except INCOMING are native to the library.

PR #147, opened 2017-06-13
users+scheme+portals: Roadwarriors users feature

### LINKED ISSUE #137: Roadwarrior users
Roadwarrior users, users that travel often and connect from external networks, force Companies to disable the [IP filter security mechanism](https://irontec.github.io/ivozprovider/en/security/index.html#authorized-company-ip-ranges).

It would be desirable if these users could be marked as roadwarriors and skip this IP check, so that their company is still protected with this IP filter.

Implementation approach:

- Users marked as roadwarrior skips the company IP check.

- These users specify a new limit: "Outgoing calls limit" (default value 3, 0 disables).

- This limit applies for both internal/external calls generated by the user (opossed to current "Maxcalls", that filters the calls that a users receives).

- Company's external max calls will apply for these roadwarrior users too.

cruzccl:
Roadwarrior users, users that travel often and connect from external networks,
force Companies to disable the IP filter security mechanism.

It would be desirable if these users could be marked as roadwarriors and skip
this IP check, so that their company is still protected with this IP filter.

The number of outgoing calls of this users must be limited to 1, 2 or 3.

This fixes #137


portals/application/configs/klear/model/Users.yaml
```diff
@@ -321,6 +321,27 @@ production:
...
+            title: "1"
+          '2': 
+            title: "2"
+          '3': 
+            title: "3"
+      info:
+        type: box
+        position: left
+        icon: help
+        text: _("Enable access from non-granted IP addresses for this user. It limits the number of outgoing calls to avoid toll-fraud. 'None' value makes outgoing calls unlimited as long as company IP policy is fulfilled.")
```
Kaian: Someone, somaday, may translate that for you.

Kaian:
I would chose kamusers as the main tag for this commit as the impact is mostly in the logic from that proxy.

portals/application/configs/klear/model/Users.yaml
```diff
@@ -321,6 +321,27 @@ production:
...
         'null': _("Company's default")
+    externalIpCalls: 
+      title: _('Calls from external IPs')
+      type: select
+      defaultValue: 0
+      source: 
+        data: inline
+        values: 
+          '0': 
+            title: _("None")
```
mmadariaga: Wouldn't "Company IP policy" or something similar be more intuitive?

Issue #49, opened 2017-04-24
recordings: Do not save them all in a unique directory


cruzccl:
year/month/day approach?

Saving them all in the same directory will lead us to trouble.

Kaian:
I though this was already done by Mappers FSO system but doesn't seem the case.

We convert from wav to mp3 in the [recordingsDir](https://github.com/irontec/ivozprovider/blob/bleeding/kamailio/recordings/application/controllers/RecordingsController.php#L138). Then we [put the file using FSO](https://github.com/irontec/ivozprovider/blob/bleeding/kamailio/recordings/application/controllers/RecordingsController.php#L192) which removes the mp3 file from that dir and moves it to the final directory in storage (_ivozprovider_model_recordings.recordedfile_) but does not create any subdir like in other cases (for example, fax files).

Are we doing something wrong?
Maybe we must first save the Recordfile and then put the file?

PR #154, opened 2017-06-14
mappers: fixed recording path issues

mmadariaga:
This PR fixes https://github.com/irontec/ivozprovider/issues/49

### DIFF
--- a/debian/ivozprovider-profile-data.install
+++ b/debian/ivozprovider-profile-data.install
@@ -1,4 +1,5 @@
 # ... FIXME ...
 profiles/data/*         /
 scripts/CDRparser.pl    /opt/irontec/ivozprovider/scripts/
+scripts/FSObase2index.sh /opt/irontec/ivozprovider/scripts/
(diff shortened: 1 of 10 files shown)


Kaian:
Could we do the same for `Locutions` table? :heart:

mmadariaga:
@Kaian You've got it

Kaian:
In order to create an script that moves existing Recordings and Locutions to their final patch, can you confirm me this logic?

```
If id < 10 dir = 0;
if id > 10 dir = each digit of id separated by '/'
```

Kaian:
Umm maybe I can come with a better logic:
```
Remove all digits but last of the id and create a folder with those splited.
If no digit has been removed (because id only has one digit) create a dir with 0.
```

mmadariaga:
@kaian You can check out the logic applied here https://github.com/irontec/klear-library/blob/chloe/Model/Fso/Adapter/StoragePathResolver/Default.php#L182

Your second approach sound good :)

PR #157, opened 2017-06-15
i18n: re-generate translation files from command line

Kaian:
Due to the fact that PoEdit behaves erratically based on the version,
the host and the current wheather, we have decided to regenerate
portal translations from command line with the hope of making the
translation process a bit less PITA.


Kaian:
Very very very far from perfect, It's just a start that requires evolving.

You can update current translations by running `make gettext` in portals dir.
You can regenerate local .mo files running `make update` although this files are no longer commited.

vslavik:
> Due to the fact that PoEdit behaves erratically based on the version,
the host and the current wheather

FWIW, Poedit actually *always* ensures consistent formatting by normalizing to GNU gettext standard formatting. (Well, unless you're using some prehistoric version, that is.)

Kaian:
Sorry I did pretend no offense there :innocent: 

The problem is not related actually related with PoEdit, but with our differents enviroments, distributions and versions of the software and what is more, different extractors configurations. That was generating differente .po files after scanning the files making the file change a lot between updates. The workaround was that all the generation was being requested to one of the developers so we decided to switch to a common way to generate the files.

**The tool is perfect as-is, the problem was our lack of consensum**. 

Again, sorry if that triggered, I was joking :+1: 

Thanks for the comment @vslavik  !!

vslavik:
> Again, sorry if that triggered

Not at all! Just meant to point out that standardizing on Poedit ≥ 1.8 can probably save a headache or two too ;)

PR #164, opened 2017-06-15
as: replace default pickup builtin feature code

### LINKED ISSUE #163: *8 as service code conflicts with Asterisk features
Even though Asterisk features are disabled, 'features show' in CLI show that there is a bind between *8 and pickup.

This internal bind makes *8 unusable as a service code.

Change this bind to a code that avoids collision with any elegible service code.

Kaian:
Default Asterisk builtin feature code for pickup is *8

This can clash with configured Service feature codes, and not knowing how to disable this feature,
we'll change it to something that won't be dialed (easily).

This allow dialplan execution for *8 services codes.

This PR fixes #163


Kaian:
This does not affect media liberation.

The builtin feature was already being used even commented.

PR #170, opened 2017-06-16
i18n: reviewed portal translations to spanish

### LINKED ISSUE #151: i18n: Translate new sections and help sections
- [x] Invoices library change

Kaian:
Generated with the upated script from #157 

General review of all translations of the portals, still lot to complete, but Its a good straring point to detect any template generation errors.

This also aims to fix #151


Kaian:
This last commits adds a new option to xgettext to ensure the generated .po translations are always sorted and removes the update timestamp of the header.

This way when a new translations is added to the .po file, the change should be just that translation.

PR #186, opened 2017-07-03
scheme: Drop incorrect unique in InvoiceTemplates

cruzccl:
The collision domain must be brand,name (it already has a unique key).


cruzccl:
This UNIQUE was removed a lot of time ago, sorry for that :(

Close and forget.

PR #194, opened 2017-07-04
recordings: fix duration and date for created recordings

Kaian:
This commit changes the way recordings duration is
calculated. Insted of using the created and modified
dates of the file, we let zend-mp3[0] library to
do the task.

We have also changed the recording date to match
the call start-time instead of using the current
timestamp. Note that this will set all on-demand
recordings within the same call with the same
timestamp.

[0] https://github.com/debach/zend-mp3.git

### DIFF
--- a/library/composer.json
+++ b/library/composer.json
@@ -6,6 +6,7 @@
         "h4cc/wkhtmltopdf-amd64": "0.12.3",
         "h4cc/wkhtmltopdf-i386": "0.12.3",
         "crada/php-apidoc":"@dev",
-        "xamin/handlebars.php": "^0.10.4"
+        "xamin/handlebars.php": "^0.10.4",
+        "debach/zend-mp3": "1.8.*"
     }
 }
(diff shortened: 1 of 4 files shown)


library/composer.json
```diff
@@ -6,6 +6,7 @@
         "h4cc/wkhtmltopdf-amd64": "0.12.3",
         "h4cc/wkhtmltopdf-i386": "0.12.3",
         "crada/php-apidoc":"@dev",
-        "xamin/handlebars.php": "^0.10.4"
+        "xamin/handlebars.php": "^0.10.4",
+        "debach/zend-mp3": "dev-master"
```
mmadariaga: Even though this library hasn't been updated in a while I would point to 1.8.1 instead of dev-master

Kaian:
```
"debach/zend-mp3": "1.8.1"
```
or should I add any special prefix ?

mmadariaga: 1.8.1 it's fine. You could use 1.8.* if you prefer, or any other "patches only" contraint as described in https://getcomposer.org/doc/articles/versions.md#writing-version-constraints

Kaian: There you go!

Issue #192, opened 2017-07-04
Outgoing DDI rules


cruzccl:
Currently when a user calls, its outgoing DDI is used as presentation number.

This feature aims to add flexibility to this behaviour, letting you use a different DDI when calling to specific destinations:

**Changes in Users:**

- Users will need an outgoingDDI as they do now.

- Users may have an OutgoingDDIRule assigned.

**New section: OutgoingDDIRules:**

- _Name_
- _Default action_: what will be done if no rule applies (Keep original DDI / Force DDI)
- _Forced DDI_: DDI used if previous setting is "Force DDI".
- None, one or multiple rules:
  - _Destination regexp_: when will this rule apply.
  - _Action_: what will this rule do (Keep original DDI / Force DDI).
  - _Forced DDI_: DDI used if previous setting is "Force DDI".
  - _Metric_ (to handle collisions).

**Usage examples**

- User that shows his/her DDI only when calling to certain destinations and shows a global DDI for remaining destinations.
  - Name: Known destinations
  - Default action: force DDI
  - Forced DDI: company global DDI
  - One rule per known destination with "keep original DDI" as action.
  - User OutgoingDDI: the one that points to this user.

- User that shows a german DDI when calling to Germany, a french one when calling to France and a generic DDI when for remaining countries.
  - Name: Country specific
  - Default action: keep original
  - One rule per country with "force DDI" as action and that country's DDI as forced value.
  - User OutgoingDDI: company global DDI.

Kaian:
Implemented in #196

Issue #89, opened 2017-05-12
kamusers: normalize numbers to/from E.164


Kaian:
The current number format flow for external calls generated by users is:

User [dialed] -> kamusers [dialed] -> PBX [e164] -> kamtrunks [e164] -> Carrier [transformed]

**_dialed_**: original number dialed by the user
**_e164_**: tranformed number from _dialed_ to _e164_ based on user country
**_tranformed_**: tranformed number from _e164_ to expected carrier format based on [transformation rules](https://irontec.github.io/ivozprovider/en/external_incoming_calls/numeric_transformations.html)

This introduces [some logics](https://github.com/irontec/ivozprovider/blob/bleeding/library/IvozProvider/Model/Users.php#L280) at PBX level to transform what user have dialed to e.164 based on a [complex regexp](https://github.com/irontec/ivozprovider/blob/bleeding/scheme/deltas/005-chile-calling-format.sql) that don't perform as good as one would guess at first sight.

There is also a [special process](https://github.com/irontec/ivozprovider/blob/bleeding/asterisk/config/dialplan/default.conf#L102) at PBX that avoids presentation of e164 to users and _guesses_ what user has originally dialed or how a number should be presented to that user to put that number again in their screen.

PBX checks ACLs, PricingPlans and Bouncing matching the number in E.164, so the original dialed number is not being used at all.

Proposal:
 - Stop using Countries regexp at PBX level to determine the real destination from what user dialed
 - Implement number transformation at Kamailio Users proxy so all PBX received and sent numbers are in E.164 regarding if it comes from a KamUsers or Kamtrunks.
 - Move the company external prefix check from PBX to kamusers
 - Update all portal fields that can be routed to an external number so it can contains Country (like DDIs do right now)

User [dialed] -> kamusers [e164] -> PBX [e164] -> kamtrunks [e164] -> Carrier [transformed]

Kaian:
Aside from the virtue of having only one numbering format at PBX, I think this makes future enhancements easier to implement.

For example, users could dial special prefix to their numbers to activate specific routing options. In this case, KamUsers will strip that prefix and fill an X-Info SIP header so that information can be transported to KamTrunks proxy ("maybe untoched").

Kaian:
Implemented in #313

PR #196, opened 2017-07-05
Outgoing ddi rules

Kaian:
Add two new tables for OutgoingDDIRules and OutgoingDDIRulePatterns.

Users can now assign a outgoingDDIRule that will determine
the final DDI presented based on the destination number
and the matching patterns in OutgoingDDIRulePatterns.

This PR implements #192


Kaian:
Should company have default outgoing ddi rules?

They have default outgoing ddi...

cruzccl:
It would be nice as it would enable something like "all the users calling to XXXXX (or to Y country) will show ZZZZZZ instead of their normal number" without having to add this rule for all the users (and new users, that will be forgotten for sure).

Collision between user rules and company rules should be solved as usual: if a user rule applies, use it. If not, evaluate company rules.

Nice feature!

library/IvozProvider/Mapper/Sql/OutgoingDDIRules.php
```diff
@@ -0,0 +1,41 @@
...
+class OutgoingDDIRules extends Raw\OutgoingDDIRules
+{
+    protected function _save(\IvozProvider\Model\Raw\OutgoingDDIRules $model,
+            $recursive = false, $useTransaction = true, $transactionTag = null, $forceInsert = false
+    )
+    {
+        $nullableFields = array(
+                "force"        => "forcedDDIId",
+        );
+        $routeType = $model->getDefaultAction();
```
Kaian: `s/routeType/defaultAction/`

library/IvozProvider/Mapper/Sql/OutgoingDDIRulesPatterns.php
```diff
@@ -0,0 +1,41 @@
...
+class OutgoingDDIRulesPatterns extends Raw\OutgoingDDIRulesPatterns
+{
+    protected function _save(\IvozProvider\Model\Raw\OutgoingDDIRulesPatterns $model,
+            $recursive = false, $useTransaction = true, $transactionTag = null, $forceInsert = false
+    )
+    {
+        $nullableFields = array(
+                "force"        => "forcedDDIId",
+        );
+        $routeType = $model->getAction();
```
Kaian: `s/routeType/action/`

Kaian:
I think this should be merged after #197 and replace the regexp field in OutgoingDDIRulePatterns with matchListId so all numeric regular expressions are in the same entity.

This is a bit less _confortable_ for company administrators as they have to also create matchlist for OutgoingRulesPatterns but will about duplicity of data and process (where the same logic of _number matches regexp_ is spread in multiples files)

Kaian:
Blocked by #201

library/IvozProvider/Model/Raw/Companies.php
```diff
@@ -3835,6 +3847,96 @@ public function getMusicOnHold($where = null, $orderBy = null, $avoidLoading = f
...
     /**
+     * Sets dependent relations OutgoingDDIRules_ibfk_1
+     *
+     * @param array $data An array of \IvozProvider\Model\Raw\OutgoingDDIRules
+     * @return \IvozProvider\Model\Raw\Companies
+     */
+    public function setOutgoingDDIRules(array $data, $deleteOrphans = false)
+    {
+        if ($deleteOrphans === true) {
+
```
mmadariaga: I think that we'll need to set some coding standards for white spaces :)

portals/application/modules/rest/controllers/OutgoingDDIRulesPatternsController.php
```diff
@@ -34,14 +34,22 @@ public function init()
      * @ApiReturn(type="object", sample="[{
      *     'id': '', 
      *     'outgoingDDIRuleId': '', 
+<<<<<<< HEAD
```
mmadariaga: Review this please

library/IvozProvider/Model/Raw/Companies.php
Kaian: But that code is autogenerated by run-generators phing task

Kaian:
I have update all source to use MatchLists instead of Regular expressions.

Also added OutgoingDDIRules at company level to make a global outgoing ddi selection (that can be overrider by user configuration).

@cruzccl 
You're example should be possible with actual implementation of outgoing ddi rules
- Create N matchlists: french, UK, ..., partners' cellphones
- Create an outgoing rule set and add one rule for each matchlist with the proper action (force french DDI, UK, and so on).

I think multiple rules will make thing more complicated, and current status covers the most common cases.

Kaian:
I must also add that this feature **does NOT work with diverted calls** until #89 is implemented.

The main reason is that all _number_ input fields of the portals doesn't have an associated country, so its value has to pass for all the _preferred_ formating logics (which I prefer to avoid).

Kaian:
Added documentation

library/IvozProvider/Model/Raw/Companies.php
mmadariaga: Thats right :sweat:

PR #197, opened 2017-07-06
portals: Enable company security by default

cruzccl:
Show both mechanisms on creation:

* Limit external calls (default value 2 in retail, 10 in vPBX)
* IP filter (disabled by default)

This way, on fast company creation, at least external calls will be limited.


Kaian:
This may require some translations.
Please, wait until #196 has been merged, rebase your changes on top of it and check if there are new strings.

Thanks!

Issue #200, opened 2017-07-09
grandstream ip phones templates


manfer:
I'm trying to create templates to provision grandstream phones but I'm not able to do so.

Grandstream phones can provision in two phases.

When the phone boots it supports DHCP option 66 and DHCP option 43 (in more recent models) to get the provision path. We can config this path manually too.

The phone will try to get the provision template from the configured path.

First it searches for  cfg000B82000000.xml file -where 000B82000000 is phone MAC- and if it doesn't find that file it searches for cfg.xml file.

So if path is `198.51.100.10/gs` it will first try to search for `http://198.51.100.10/gs/cfg000B82000000.xml` and if that doesn't exists it will try `http://198.51.100.10/gs/cfg.xml`. These are examples for http provision but there is a config value to make it search on http, https or tftp.

The first problem I found is that the phone needs xml files not plain text files, but that can be easily solved as we can write php code on the template field.

So the generic file is easy. Something like this.

```php
<?php 
header('Content-Type: application/xml');
?><?xml version="1.0" encoding="UTF-8" ?> 
<gs_provision version="1">
  <config version="1">
    <P212>2</P212>
    <P237><?php echo $_SERVER['SERVER_NAME']; ?>:443/provision/gxp2130</P237>
    <P8387>1</P8387>
  </config>
</gs_provision>
```

The code for the specific file is easy to code too. It would be something like this:

```php
<?php 
header('Content-Type: application/xml');
?><?xml version="1.0" encoding="UTF-8" ?> 
<gs_provision version="1">
  <mac><?php echo $this->terminal->getMac(); ?></mac>
  <config version="1">
    <P270><?php echo $this->terminal->getName(); ?></P270>
    <P47><?php echo $this->company->getDomainUsers(); ?></P47>
    <P35><?php echo $this->user->getName(); ?></P35>
    <P36><?php echo $this->user->getName(); ?></P36>
    <P34><?php echo $this->user->getPassword(); ?></P34>
    <P3><?php echo $this->user->getName(); ?></P3>
    <P33>*93</P33>
  </config>
</gs_provision>
```

The problem is that I don't know how can I manage to make ivozprovider to reply to request to cfg{MAC}.xml files as I can not edit the path in the specific file.

It is a fixed value `https://{PROVISIONING_URL}:{PORT_IN_GENERIC_FILE}/provision/{OPTIONAL_SUBPATHS}/`

[...]

PR #201, opened 2017-07-10
ExternalCall Filters whitelists and blacklists

Kaian:
This PR includes a new object call Match Lists that can be used to group specific Numbers or Regular Expressions. 

These lists could be used in multiples configurations, but commits below only implements Black and White lists for External Call filters, aiming to implement #27 and #28 

Another examples where this could be used are #192  and future conditional routing options.

PR #202, opened 2017-07-10
provision: customizable specific provision routes

mmadariaga:
Make speficic provision file name editable as requested in https://github.com/irontec/ivozprovider/issues/200

This is the proposed flow:

1 - Extract any posible mac from the url
2 - Search thoose values into Terminals table
3 - Double check against specificUrlPattern in TerminalModels table

This PR aims to close https://github.com/irontec/ivozprovider/issues/200

Apart from custom file formats, the following mac formats should work from now on:
- https://test/provision/xxXXxxXXxxXX.cfg
- https://test/provision/xx:xx:xx:xx:xx:xx.cfg
- https://test/provision/xx-xx-xx-xx-xx-xx.cfg

### DIFF
--- a/library/IvozProvider/Mapper/Sql/TerminalModels.php
+++ b/library/IvozProvider/Mapper/Sql/TerminalModels.php
@@ -20,7 +20,8 @@
 namespace IvozProvider\Mapper\Sql;
 class TerminalModels extends Raw\TerminalModels
 {
-    public function save(\IvozProvider\Model\Raw\TerminalModels $model, $forceInsert = false){
+    public function save(\IvozProvider\Model\Raw\TerminalModels $model, $forceInsert = false)
+    {
         $genericMustChange = $model->hasChange("genericTemplate");
         $specificMustChange = $model->hasChange("specificTemplate");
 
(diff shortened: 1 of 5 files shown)


Kaian:
#196 has delta 063 and #201 has delta 064.

Do you want me to update those PR deltas or should I block this after they're merged? They will still be unmerged for a couple of days.

library/IvozProvider/Mapper/Sql/TerminalModels.php
```diff
@@ -34,19 +35,19 @@ public function save(\IvozProvider\Model\Raw\TerminalModels $model, $forceInsert
...
+            try {
                 $template = $model->getGenericTemplate();
                 $this->createFolder($route);
                 $file = "generic.phtml";
                 $this->saveFiles($file, $route, $template);
-            }
-            catch (\Exception $e){
+
+            } catch (\Exception $e) {
                 throw $e;
```
Kaian: This is the default Exception behaviour.

portals/application/modules/provision/controllers/IndexController.php
```diff
@@ -118,28 +118,63 @@ protected function _searchGenericPattern(\IvozProvider\Mapper\Sql\TerminalModels
...
+             * @var IvozProvider\Model\TerminalModels $terminalModel
+             */
+            $terminalModel = $candidate->getTerminalModel();
+            if (!$terminalModel) {
+                continue;
+            }
 
-        if (!$terminal) {
-            return null;
+            $especificUrl = $this->extractFileNameWithoutExtension(
```
Kaian: _Specific_ or _Especific_ here?

portals/application/modules/provision/controllers/IndexController.php
```diff
@@ -118,28 +118,63 @@ protected function _searchGenericPattern(\IvozProvider\Mapper\Sql\TerminalModels
...
+
+            if (strtolower($fixedEspecificUrl) === strtolower($fileName)) {
+                return $candidate;
+            }
+        }
+
+        return null;
+    }
+
+    protected function extractFileNameWithoutExtension($route)
```
Kaian: We did this before for multimedia worker using pathinfo. It works with strings (the file doesn't have to exist), check it out [here](https://github.com/irontec/ivozprovider/blob/bleeding/portals/application/workers/MultimediaWorker.php#L50)

Kaian:
I will grant this PR the privilege of delta 63, so It can be backported to oasis easily.

I will mark #196 and #201 as blocked by this.

Thanks!!

portals/application/modules/provision/controllers/IndexController.php
mmadariaga: :sweat_smile:

portals/application/modules/provision/controllers/IndexController.php
mmadariaga: Thanks. I'll keep the method changing the logic

PR #212, opened 2017-07-18
portals: fixed sql condition

mmadariaga:
Fixed sql condition where calls were not being filtered by company properly


Kaian:
Nice, this should be merged into stable

PR #214, opened 2017-07-20
portals: make peering contract visible in billable calls

mmadariaga:
Make peering contract visible in billable calls


Kaian:
I think this is also required for KamAccCdrsBrandList.yaml  :thinking:

PR #217, opened 2017-07-20
GS Wave QR code provision

manfer:
This adds a feature to insert a field in users configuration form to enable a QR Code in the users portal to provision GS Wave.

GS Wave is a free softphone app available for android and IOS devices

### DIFF
--- a/portals/application/modules/userweb/controllers/AuthController.php
+++ b/portals/application/modules/userweb/controllers/AuthController.php
@@ -55,6 +55,30 @@ class Userweb_AuthController extends Iron_Controller_Rest_BaseController
             }
         }
 
+        $gsQRCode = $user->getGsQRCode();
+        $result['gsQRCode'] = $gsQRCode;
+
+        if ($gsQRCode) {
+
+            if (!empty($terminal)) {
+                $terminalPassword = $terminal->getPassword();
+                $result['terminalName'] = $terminalName;
+                $result['terminalPassword'] = $terminalPassword;
+            }
+
+            $extension = $user->getExtension();
+
+            if (!empty($extension)) {
+                $extensionNumber = $extension->getNumber();
+                $result['extensionNumber'] = $extensionNumber;
+            }
+
+            $result['companyDomain'] = $user->getCompany()->getDomain();
+
+            $result['voiceMail'] = '*' . $user->getCompany()->getServiceCode('Voicemail');
+
+        }
+
         $result['success'] = true;
         $result['message'] = 'Login Correcto';
         $result['userName'] = $user->getFullName();
--- a/userweb/app/modules/commonView/headerCtrl.js
+++ b/userweb/app/modules/commonView/headerCtrl.js
@@ -46,7 +46,41 @@ angular
             $scope.userAgent = localStorage.getItem('userAgent');
             $scope.ipRegistered = localStorage.getItem('ipRegistered');
         }
-        
+
+        var gsQRCode = localStorage.getItem('gsQRCode');
+
+        if (gsQRCode == 1) {
+
+            var terminalName = localStorage.getItem('terminalName');
+            var terminalPassword = localStorage.getItem('terminalPassword');
+            var extensionNumber = localStorage.getItem('extensionNumber');
+            var companyName = localStorage.getItem('companyName');
+            var companyDomain = localStorage.getItem('companyDomain');
+            var userName = localStorage.getItem('userName');
+            var voiceMail = localStorage.getItem('voiceMail');
+
+            $scope.gsQRCode = true;
+
+
+            $scope.QRCode = '<?xml version="1.0" encoding="utf-8"?>'
+                + '<AccountConfig version="1"><Account>'
+                + '<RegisterServer>' + companyDomain + '</RegisterServer>'
+                + '<UserID>' + terminalName + '</UserID>'
+                + '<AuthID>' + terminalName + '</AuthID>'
+                + '<AuthPass>' + terminalPassword + '</AuthPass>'
+                + '<AccountName>' + extensionNumber + ' ' + companyName + '</AccountName>'
+                + '<DisplayName>' + userName + '</DisplayName>'
+                + '<Voicemail>' + voiceMail + '</Voicemail>'
+                + '</Account>'
+                + '</AccountConfig>';
+
+        } else {
+
+            $scope.gsQRCode = false;
+            $scope.QRCode = '';
+
+        }
+
     };
     
     loadData();
--- a/userweb/app/styles/main.scss
+++ b/userweb/app/styles/main.scss
@@ -261,6 +261,39 @@ table {
     }
 }
 
+/**
+ * QR Code
+ */
+#qrcode {
+    position: relative;
+}
+
+#qrcode div {
+    position: absolute;
+    left: 0;
+    right: 0;
+    width: 224px;
+    margin: 0 auto;
+    color: rgba(0,0,0,.8);
+    font-size: 10pt;
+    font-style: italic;
+    top: -9999px;
+    border: 6px solid rgba(0,0,0,.15);
+    border-radius: 6px;
+    padding: 6px;
+    background-color: rgba(255,255,255,.95);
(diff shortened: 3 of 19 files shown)


portals/application/modules/userweb/controllers/AuthController.php
```diff
@@ -33,12 +34,25 @@ public function postAction()
...
+        $company = $user->getCompany();
 
         if (!empty($terminal)) {
 
             $dbAdapter = Zend_Db_Table::getDefaultAdapter();
             $terminalName = $terminal->getName();
 
+            $gsQRCodeEnabled = $user->getGsQRCode();
+            $hasFeatureQRCode = $company->hasFeature(Features::QRCODE);
+            $gsQRCode = ($gsQRCodeEnabled AND $hasFeatureQRCode) ? 1 : 0;
```
mmadariaga: In order to keep current code conventions, could you replace 'AND' by '&&' please?

portals/application/modules/userweb/controllers/AuthController.php
```diff
@@ -55,11 +69,24 @@ public function postAction()
...
+            }
+        }
+
         $result['success'] = true;
         $result['message'] = 'Login Correcto';
         $result['userName'] = $user->getFullName();
         $result['companyName'] = $user->getCompany()->getName();
 
+        if ($gsQRCode) {
+            $result['companyDomain'] = $user->getCompany()->getDomain();
```
mmadariaga: I would combine those three conditions "if ($jsQRCode)" into a single one. What do you think about that?

mmadariaga:
Good contribution by the way :)

Kaian:
Hi @manfer 

First of all, thanks for taking the time and having the pacience to dig in the code and learn the development process that has been required to implement this feature (that works like a charm indeed). For my point of view, that's something admirable!

All Pull Requests (PR) of the project must be reviewed and approved by at least two core developers (same applies even to our PRs), so I'll give my two cents here. The code is yours so your oppinion counts as much as ours for these:

1. QR position in userweb
The QR in userweb is used only during the initial provisioning of the gwsave, so the rest of the time doesn't provide any information in the portal. Our testing QR was quite big, and at first the location was ok, easy to find, but after some time we thing it will be better if it's located in "My Account" section. Another alternative would be to add something toggle its visibility in its current position.

2. QR Feature
We're glad that you have taken the time to even learn how features work and make a new one. But I think features should be more generalistic and will affect more than a single field. If could make sense if we have some fields or processes related with Grandstream and we want to enable/disable them for some companies. For me it's perfectly fien if  qrcode field in Users' edit screen is visible for all the companies. Maybe you feel the need to hide it for some reason.

Any other thing that we came accross we'll provide inline comments next to the code.

Good work, and thanks for the contribution!

userweb/app/modules/commonView/headerCtrl.js
```diff
@@ -46,7 +46,23 @@ angular
...
+        if (gsQRCode == 1) {
+            var terminalName = localStorage.getItem('terminalName');
+            var terminalPassword = localStorage.getItem('terminalPassword');
+            var extensionNumber = localStorage.getItem('extensionNumber');
+            var companyName = localStorage.getItem('companyName');
+            var companyDomain = localStorage.getItem('companyDomain');
+            var userName = localStorage.getItem('userName');
+
+            $scope.gsQRCode = true;
+            $scope.QRCode = '<?xml version="1.0" encoding="utf-8"?><AccountConfig version="1"><Account><RegisterServer>' + companyDomain + '</RegisterServer><UserID>' + terminalName + '</UserID><AuthID>' + terminalName + '</AuthID><AuthPass>' + terminalPassword + '</AuthPass><AccountName>' + extensionNumber + ' ' + companyName + '</AccountName><DisplayName>' + userName + '</DisplayName><Voicemail>*93</Voicemail></Account></AccountConfig>';
```
Kaian: This could improve readability with a multine string, what about:

```            
$scope.QRCode = '<?xml version="1.0" encoding="utf-8"?>'
            + '<AccountConfig version="1"><Account>'
                + '<RegisterServer>' + companyDomain + '</RegisterServer>'
                + '<UserID>' + terminalName + '</UserID>'
                + '<AuthID>' + terminalName + '</AuthID>'
                + '<AuthPass>' + terminalPassword + '</AuthPass>'
                + '<AccountName>' + extensionNumber + ' ' + companyName + '</AccountName>'
                + '<DisplayName>' + userName + '</DisplayName>'
                + '<Voicemail>*93</Voicemail>'
            + '</Account></AccountConfig>';
```

Kaian: This also assumes the Voiceimail code is `*93` which can be configured per company basis.
We can create a Company method called getCompanyServiceCode in [its model](https://github.com/irontec/ivozprovider/blob/bleeding/library/IvozProvider/Model/Companies.php#L352) that returns the code related to a service so you can use `$voicemailExten = '*' . $company->getCompanyServiceCode('Voicemail');`

manfer:
I will work on changes this weekend. I will read more carefully but at first glance all suggestions looks ok to me.

Till now I was just rebasing and doing forced push. I bet new commits would be better from now on.

So just a question. Is it ok if I do changes with interactive rebases and force push (lost history) or you prefer new commits without breaking the branch history?

Kaian:
For feature branches, force pushes are allowed, unless other features are based on your branch or multiple developers are working in the same branch.

For this PR branch, you can safely force push, the PR will be updated automatically with the new commit history.

stable (oasis) and testing (bleeding) are protected branches and forced pushes are not allowed.

Have a nice weekend!

Kaian:
Blocked by #196 (scheme delta numbering)

manfer:
I have pushed some changes. Waiting your opinion. Thanks.

mmadariaga:
Looks good to me

userweb/app/styles/main.scss
```diff
@@ -261,6 +261,36 @@ table {
...
+    border-radius: 6px;
+    padding: 6px;
+    background-color: rgba(255,255,255,.95);
+    -webkit-box-shadow: 0 0 6px rgba(132, 132, 132, .75);
+    -moz-box-shadow: 0 0 6px rgba(132, 132, 132, .75);
+    box-shadow: 0 0 6px rgba(132, 132, 132, .75);
+    -webkit-border-radius: 4px;
+    -moz-border-radius: 4px;
+}
+
```
Kaian: After testing this, the overlayed image flickers a bit when the cursor is over.

We have replaced this CSS to a simple _resize_ behaviour. Please, give it a try if you have the time.

```/**
 * QR Code
 */
#qrcode {
    position: relative;
}

#qrcode:hover div {
    display:block!important;
}

#qrcode:hover > qr {
    display:none;
}

#qrcode div {
    display: none;
    color: rgba(0,0,0,.8);
    font-size: 10pt;
    font-style: italic;
}
```

What do you think?

userweb/bower.json
```diff
@@ -20,7 +20,8 @@
     "angularjs-datepicker": "~0.2.4",
     "angular-i18n": "~1.3.15",
     "moment": "~2.10.2",
-    "crypto-js": "~3.1.4"
+    "crypto-js": "~3.1.4",
+    "angular-qr": "latest"
```
Kaian: We prefer avoid ultra dynamic versions for bower and node modules.

I would replace _latest_ with `~0.2.0` or `0.2.0` which seems the latest [angular-qr release](https://github.com/janantala/angular-qr/releases)

userweb/app/styles/main.scss
manfer: I tested on another resolution and the popup was not over the small QR code. That's why it flickers. Those should not overlap.

The idea is to have the hover only on the small qrcode. I will rework that CSS.

manfer:
CSS reviewed. Now it should work correctly.

angular-qr release version added on bower.json.

cruzccl:
Really good contribution @manfer :+1: :clap: 

Thank you very much!

Kaian:
I'm ready to merge this, would you mind to do a final rebase over latest `bleeding` changes and update the PR?

I want to keep your original branch commits during the merge :)

Thanks!

manfer:
ready. Thanks.

PR #219, opened 2017-07-24
portals: logo size limited by height

manfer:
There is nothing that prevents the logo from being wider. There is a lot of white space to the right.

Wouldn't it be better to limit the logo size by height?

I decided to set it to 96px which is your actual logo height.

### DIFF
--- a/portals/application/assets/css/global.css
+++ b/portals/application/assets/css/global.css
@@ -1,5 +1,5 @@
 #applicationLogo img {
-    max-width: 250px;
+    max-height: 80px;
     margin-top: 10px;
 }
 


mmadariaga:
Could you link us an image so that we can reproduce this issue? Thanks

manfer:
Any wide logo (more than 250px) will show the problem. The wider the logo the worst. Because the max-width forces a proportional shrink of image height on logos wider than 250px (max-width value). The shrinked image does not look good because has a lot of bottom wide margin.

If the image has small text somewhere this is even worst.

Some images to test.
https://www.brenau.edu/logos/BU_black_gold_wide.jpg
https://static1.squarespace.com/static/5321daaae4b0740a41873481/t/53a4a0e9e4b0b0e684bb7155/1403298025540/Puppet_logo_3_pan2587.jpg

I changed CSS limiting by height to 96px because your logo is 96px high though I think 86px limit would be better. Because more than 86px height pushes left menu panel down.

Take in mind that the images are never scaled up. We are only configuring a limit to shrink images if the admin uploads a large one.

mmadariaga:
Sorry, I must be missing something here but I haven't been able to reproduce it. 

attempt3

attempt1
attempt2

manfer:
This is just a matter of aesthetics. There is a lot of horizontal space that can be used for the logo. Why the 250px width limit?

Why not limit logo size by height instead? I decided to use 96px height because that's your sample logo height but, considering the head div has almost exactly 100px and that the logo image has a 10px top margin, the best looking max-height would be 80px, so the remaining bottom margin is another 10px.

# max-height 80px
max-height

# max-width 250px
max-width

Which one looks better?

This is just to shrink the image when the admin uploads a large image. A good admin should upload the image with the resolution he wants. But if his logo is a wide logo he will hit the arbitrary (unless there is a good reason for this) 250px imposed width limit.

manfer:
I have changed the max-height proposed value to 80px as your sample logo looks good shrinked to 80px height.

Kaian:
I also like this!

As your contribution grows, I will like to mention that we tag commits with some prefix so we can determine where the change are being applied at first glance.

There is some documentation here: [es](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/es/commits.md) [en](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/en/commits.md)

If you don't mind rewording the commit to: `portals: logo size limited by height` we will include it in 1.5.0

Thanks!

PR #221, opened 2017-07-24
doc: queue and huntgroup strategies

### LINKED ISSUE #204: doc: Queue strategies
Add some documentation explaining different call distribution strategies supported by queues.

Kaian:
I have updated portals fields and documentation for both Queues and Huntgroups so the description of each available strategy is documented in the last one. 

As always, some translation have changed it's format :disappointed: 

portals: Removed strategy text from queue strategy field
portals: Added hint to huntgroup strategy field
doc: added queue strategy explanations
i18n: unified labels and translations of both fields

I have also removed AstQueue and AstQueueMembers portals files as they don't have its own portal section.

This aims to fix #204


portals/application/languages/es_ES/es_ES.po
```diff
@@ -2820,6 +2781,9 @@ msgstr "Loc. Error"
 msgid "Success locution id"
 msgstr "Loc. Éxito"
 
+msgid "Timeout"
+msgstr "Timeout"
```
mmadariaga: Is this a missing translation or is it intended?

Kaian: That was detected and generated as new, but I have no idea why it's the first time it appears tbh

PR #230, opened 2017-07-26
portals: add register status icon for Terminals, Friends and Accounts

Kaian:
The icon was already present in Terminals list, so I have recoded the TerminalStatus ghost class to handle Friends and Retails too.

I have revamped the terminalStatus field to status and hide them in List screens.

Retail Accounts and Friends that has direct connectivity enabled (so no register entry is present in locations table) will always display an arrow instead of register information.


library/IvozProvider/Klear/Ghost/RegisterStatus.php
```diff
@@ -0,0 +1,80 @@
...
+        );
+
+        // Try to find registry entry
+        $locationMapper = new \IvozProvider\Mapper\Sql\KamUsersLocation;
+        $location = $locationMapper->fetchOne(implode(' AND ', $where), "expires DESC");
+
+        // Draw a green tick if found, or red cross if not
+        if (!is_null($location)) {
+            return '<span class="ui-silk inline ui-silk-tick" title="Registered until ' . $location->getExpires() . '"></span>';
+        } else {
```
mmadariaga: You can remove else condition here :wink:

Kaian:
Updated with your suggestions!

PR #244, opened 2017-08-01
scheme: remove endpoint subscribe context field

Kaian:
In asterisk 13.17.0 subscribe context is already present as an endpoint field but declared with an underscore.

This PR updates IvozProvider endpoints table so no custom patch is longer required.


scheme/deltas/068-asterisk-subscribecontext.sql
```diff
@@ -0,0 +1,6 @@
+-- Since asterisk 13.17.0 subcribe context is already and endpoint field
+-- but field is defined with underscore
+
+ALTER TABLE `ast_ps_endpoints` ADD `subscribe_context` varchar(40) NOT NULL DEFAULT 'default' AFTER `subscribecontext`;
```
cruzccl: What about renaming it with CHANGE instead of all this?
```
ALTER TABLE `ast_ps_endpoints` CHANGE `subscribecontext` `subscribe_context` varchar(40) NOT NULL DEFAULT 'default' AFTER `subscribecontext`;
```

Issue #249, opened 2017-08-02
Conditional route based on time of the day


cruzccl:
Currently IvozProvider only supports 3 logics for external incoming calls:

- Holiday behaviour (January 1st).
- Out of schedule behaviour (tuesday at 2 AM).
- General logic (calls on working hours).

It would be nice if more time conditions could be added:

- Monday-Thursday from 8 to 14: call 1001.
- Monday-Thursday from 15 to 19: call to 1002.
- Friday from 8 to 15: call to 1003.

Issue #248, opened 2017-08-02
Conditional routing based on call origin


cruzccl:
Since IvozProvider 1.5, numbers can be grouped in matchlists.

It would be nice if different call logic could be configured when any of these groups matches in incoming external calls.

**Examples:**

- Calls from numbers matching my group "Germany" (that has all ^49 numbers), go to queue 'Germany'.

- Calls from numbers matching my group "IT workers cell phones", go to queue 'IT'.

- Remaining, follow the general logic.

PR #256, opened 2017-08-10
Conditional routes based on schedules and origin

Kaian:
New tables for conditional routings

- ConditionalRoutes:
    * Main entity that can now be routed from DDIs and Extensions
- ConditionalRoutesConditions:
    * Each ConditionalRoutes can have N ConditionalRoutesConditions with
      special treatments based on the condition type
- ConditionalRoutesConditionsRelMatchLists
    * For conditions of type 'origin' this table store the relation between
      the condition and matching lists.
- ConditionalRoutesConditionsRelSchedules
    * For conditions of type 'schedule' this table store the relation between
      the condition and matching schedules.

This also updates the routable options of DDIs and Extensions.

This PR implements #248 and #249


Kaian:
TODO: Documentation

Kaian:
Updated PR with following changes:

 - Conditional Routes now route to all previous plus ConferenceRoom and Extension
 - Added new conditional match with Calendar Holidays
 - Conditional Routes now has no type: All Origin,Schedule and Calendar matches can be combined in a single condition
 - Added Routable option in IVR Custom entries to route directly using Conditional route

PR #261, opened 2017-08-11
doc: Errors in pricingPlan CSV format help

### LINKED ISSUE #257: doc: wrong decimal tip
On pricing plans documentation we can read

> Floating number must be quoted with double quotes and use a comma as decimal separator.

But the sample CSV file provided uses dot as decimal separator and it is correctly imported.

I did not try with a CSV file with commas as decimal separator so I don't know if both comma and dot works as separator or just dot.

The doc is not accurate any way and should be fixed.

Same on spanish doc.

cruzccl:
Fixed some discrepancies between documentation, helper section and example CSV.

This commit fixes #257


manfer:
I suggest using dot or point but not period to refer to the decimal separator.

https://jakubmarian.com/dot-period-full-stop-and-point-in-english/

cruzccl:
Thank you @manfer , I have already changed period with point :)

PR #273, opened 2017-08-31
kamusers: Remove redis and redis modules

cruzccl:
Redis was included as an experimental approach to have some realtime
statistics (total calls, calls per company, etc.).

This approach has been discarded and this commits tries to erase all
references to it.


extra/simple-cdd/package.list
```diff
@@ -1,4 +1,4 @@
-ivozprovider
+evozprovider
```
Kaian: :thinking:

Issue #292, opened 2017-09-22
Conflict in Companies database between vpbx and retail type


Guigui37:
If I add a vpbx company with the same name as a retail company, the WebUI gives us the following error: "Error saving record"
Portal logs show the following:
>Sep 22 14:34:40 voip-portal-2 ivozprovider[25506]: 2017/09/22 - 16:34:40 ERR (3): Exception encountered while attempting to save IvozProvider\Mapper\Sql\Companies with an empty primary key  Exception: Mysqli statement execute error : Duplicate entry 'test-1' for key 'nameBrand'
[...]

PR #297, opened 2017-09-24
portals: companies name constraint

manfer:
This aims to solve https://github.com/irontec/ivozprovider/issues/292 changing the constraint to allow same name for vpbx companies and retail companies and catching the exception on saving companies.

### DIFF
--- a/library/IvozProvider/Mapper/Sql/Companies.php
+++ b/library/IvozProvider/Mapper/Sql/Companies.php
@@ -22,6 +22,14 @@ use IvozProvider\Gearmand\Jobs\Xmlrpc;
 
 class Companies extends Raw\Companies
 {
+
+    /*
+     * Mysql error code list:
+     * https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html
+     */
+    const MYSQL_ERROR_DUPLICATE_ENTRY = 1062;
+    const UNIQUE_NAME_CONSTRAINT = 'nameBrand';
+
     protected $_model;
     protected $_recursive;
     protected function _save(\IvozProvider\Model\Raw\Companies $model,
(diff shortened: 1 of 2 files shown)


scheme/deltas/071-companies-name-constraint.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX `nameBrand` ON `Companies`; 
+CREATE UNIQUE INDEX `nameBrandType` ON `Companies`(`name`,`brandId`,`type`);
```
mmadariaga: Even if this change may make sense, I would not change the current unique key as it comes with some drawbacks: Currently, the company name is the only way to distinct them in emulation selector. I cannot think of any other issue right now.

Because of that, this would require some frontend work as well in order to be accepted.

manfer: If I understand correctly what is the emulation selector I don't think any other change is needed.

After this change in the emulation selector, if I understand correctly it is the dropdown menu to select the company you want to emulate, if there are a vpbx and a retail with same name they both appear correctly in the dropdown. A vpbx icon prepends the name of the vpbx and a retail icon prepends the name of the retail.

When you select the vpbx the system correctly emulates the vpbx. When you select the retail the system correctly emulates the retail account.

Not sure if this is what you mean.

mmadariaga: My mistake. I'm sorry, I should have taken some time to check this out beforehand. 

It turns out that emulation selector is the only select combo with icons. The issue would be applicable to the rest of the company selectors: Outgoing routing and invoices for instance.

manfer: I see now those other places. Thanks.

In my opinion it is better to disallow same name on retail and vpbx then. So I will remove the delta and change the captured constraint in Companies.php.

What's your opinion? Preserve the (name,brandId) constraint or change the code of those other company selectors in outgoing routing, invoices, ...?

mmadariaga: Adding the capability to show an icon into select options could take some time, so I would preserve the constraint too. 

Although showing country flags and stuff like that has been on our wishlist for some time now, we will be able to find some time to implement it eventually. We could take back this topic then.

Thanks for your contribution by the way :)

manfer:
Pushed changes that make the company name unique by brand. A vpbx and retail client on same brand can't have same name.

The constraint is catched on company save and shows the corresponding error popup.

PR #300, opened 2017-09-29
portals: no retail account creation if no brand sip domain

manfer:
Disallow creation of retail account if brand has not a SIP domain

### DIFF
--- a/library/IvozProvider/Mapper/Sql/RetailAccounts.php
+++ b/library/IvozProvider/Mapper/Sql/RetailAccounts.php
@@ -27,6 +27,11 @@ class RetailAccounts extends Raw\RetailAccounts
     protected function _save(\IvozProvider\Model\Raw\RetailAccounts $model,
         $recursive = false, $useTransaction = true, $transactionTag = null, $forceInsert = false
     ) {
+        // Check if brand domain is empty
+        if ($model->getCompany()->getBrand()->getDomainUsers() === '') {
+            throw new \Exception('Brand must have a SIP domain', 2204, $e);
+        }
+
         // Set account domain to its brand domain
         $model->setDomain($model->getCompany()->getBrand()->getDomainUsers());
 
--- a/portals/application/configs/klear/errors.yaml
+++ b/portals/application/configs/klear/errors.yaml
@@ -27,6 +27,7 @@ production:
     2001: _("SIP Domain can't be empty")
     2201: _("Email already in use")
     2203: _("Name already in use")
+    2204: _("Brand must have a SIP domain")
   connectorErrors:
     9001: _("No Application Server available.")
 staging:


Kaian:
This check is ok, but you're displaying an error to a company/retail admin that refers to a Brand missconfiguration. I think they shouldn't be aware of Brand existance at all, and they can not fix the problem.

The error itself is enabling Retail Feature in Brand while leaving sip domain, empty so I think the correct place for checking this will be the Brand mapper, after saving. I not sure I we can also change domain field visibility in Brand configuration screen if Features select has Retail enabled. It's better to have an extrra mandatory field rather than an error after saving.

Thanks @manfer !

Kaian:
Oh, Retail administations can not create new accounts, my bad!

Still think this should be checked where the source of the problem is.

mmadariaga:
I agree with @Kaian. We could ensure that domain field is never updated to an empty value that way.

manfer:
I agree. The logic should be on Brands creation.

- If Retail Feature is enable, sip domain must be visible and required.
- If Retail Feature is disable, sip domain must be invisible and not required.

Kaian:
I know how to do this with fields with static values (like the Route Type selectbox of a lot of screens), but not sure how to do it when values are loaded from a mapper (like Features multiselectbox).

According to [klear documentation (es)](http://irontec.github.io/klear-tutorial/yaml/fields/select.html) this could be possible, but we have to do some research :sweat_smile:

Kaian:
Although documentation mentions select, not multiselect :-/

mmadariaga:
VisualFilter attribute is handled only by select fields. Let me try to implement it

mmadariaga:
There is no documentation yet but the functionality has been already implemented  in https://github.com/irontec/klearMatrix/commit/36f3b20ea1293990b5505c99e64e3c23ad456f93. 

Feel free to apply it or just throw the exception and we'll take it from that point. An example below:
>     FeaturesRelBrands:
>      title: _('Features')
>      type: multiselect
>      source:
>        data: mapper
>        config:
>          extraDataAttributes:
>            featureId: id
>          relationMapper: \IvozProvider\Mapper\Sql\FeaturesRelBrands
>          relatedMapperName: \IvozProvider\Mapper\Sql\Features
>          relationProperty: Feature
>          relatedFieldName:
>            fields:
>              - name
>            template: '%name%'
>          visualFilter:
>            iden:
>               retail:
>               toggle:
>                 - domain_users
>             recordings:
>               toggle:
>                 - recordingsLimitMB
>                 - recordingsDiskUsage
>                 - recordingsLimitEmail
>             billing:
>               toggle:
>                 - postalAddress
>                 - postalCode
>                 - town
>                 - province
>                 - country
>                 - registryData
>         order: name

manfer:
I have not been able to make it work so could you take this one from here? If you want you can close this pull request and open a new one.

Thanks.

mmadariaga:
Sure, Thanks

mmadariaga:
Closed in favour of https://github.com/irontec/ivozprovider/pull/307

PR #301, opened 2017-09-29
portals: show domain on retail accounts list

manfer:
Show the SIP domain on retail accounts list


Kaian:
I'm ok with this. If we change this screen, we should also change [Brand Retail  Account view](https://github.com/irontec/ivozprovider/blob/oasis/portals/application/configs/klear/BrandRetailAccountsList.yaml)

Thanks!

manfer:
I edited the commit to change it too in Brand Retail Accounts view

PR #303, opened 2017-09-29
portals: change retail accounts domain length

manfer:
Brand SIP domain is a 255 length varchar so retail accounts domain must be too.


Kaian:
All domains should have a max lenght of 190 in the platform according to [this delta](https://github.com/irontec/ivozprovider/blob/oasis/scheme/deltas/041-general-review.sql).
We should change Brand's domains field and also limit it's max lenght in the screen.

Thanks for pointing this out!

manfer:
I created other pull request
https://github.com/irontec/ivozprovider/pull/304

This one can be closed

Kaian:
Good, thanks for your hard work :) :+1:

PR #304, opened 2017-10-02
portals: fix domain length

manfer:
Domain length must be 190. This aims to fix some places where it is set to 255.

### DIFF
--- a/portals/application/configs/klear/model/Terminals.yaml
+++ b/portals/application/configs/klear/model/Terminals.yaml
@@ -34,7 +34,7 @@ production:
     domain:
       title: _('Domain')
       type: text
-      maxLength: 255
+      maxLength: 190
     disallow:
       title: _('Disallowed audio codecs')
       type: text
--- a/portals/application/configs/klear/model/Brands.yaml
+++ b/portals/application/configs/klear/model/Brands.yaml
@@ -17,7 +17,7 @@ production:
       title: _('SIP domain')
       type: text
       trim: both
-      maxLength: 255
+      maxLength: 190
     defaultTimezoneId:
       title: _('Default timezone')
       type: select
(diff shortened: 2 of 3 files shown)


portals/application/configs/klear/model/Terminals.yaml
```diff
@@ -31,10 +31,6 @@ production:
         icon: help
         text: _("Allowed characters: a-z, A-Z, 0-9, underscore and '*'")
         label: _("Need help?")
-    domain:
-      title: _('Domain')
-      type: text
-      maxLength: 255
     disallow:
```
Kaian: Domain is a field in Terminals list screen, can we remove the entire field configuration?

manfer:
In list but not in model if I'm not wrong. Friends for example does not have the domain field already. The domain is inherited from company but we don't fill that field on terminals creation.

portals/application/configs/klear/model/Terminals.yaml
manfer: In list but not in model if I'm not wrong. Friends for example does not have the domain field already. The domain is inherited from company but we don't fill that field on terminals creation.

manfer: In fact it could be totally dropped from database. As you can get that value from the company itself. But I understood that was by design for some performance reason. Otherwise it would be a lot better to drop domain from Terminals, Friends and Retail Accounts.

Kaian: Afaik fields are there because it's easier for kamailio to query that tables directly when checking credentials. It's a bit PITA because when a domain changes we have to update a lot of places..

Although not used in the edit screen, Domain field is present in the list screen and the text from the header of the table is taken from the model (Domain for english, Dominio for spanish). We could remove maxLength, I think that's not used at all for listing.

manfer: Understood. I insert it again. I include maxLength too with value 190 just in case.

PR #310, opened 2017-10-11
kam: Dispatch doing hash for companies with Queues

cruzccl:
We haven't found a proper way to share queue states between Asterisk
boxes yet. That's why Companies with Queues should use only one AS.

This is currently achieved assigning the company to an AS in Companies section,
but no failover will occur if that AS is not reachable.

This issue aims to add logic that changes dispatching algorithm for Companies
without an static assignation following these rules:

- If it has queues, choose AS hashing over bXcY value (X brandId, Y companyId).

- If no queues, continue using round-robin algorithm.

In both cases, if chosen AS is not responding, another one is chosen.

It is something like "choose always the same AS as long as it is alive" for
companies with queues.


Guigui37:
I agree with @Kaian , better that current status, but not a final solution.

Thanks for your work :-)

cruzccl:
Can you take a look please to the new approach? :)

Thanks!

Issue #100, opened 2017-05-19
Evaluate: standarization to E.164 with plus sign


Kaian:
While it seems optional, standarizing to E.164 prefixing the Country Calling code with a plus sign will allow us to determine if the number is external and should require any kind of transformation.

This will also improve the readability of UI where most users are used to see the CC prefixed (the country select boxes already display it).

We should evaluate this taking into account the drawback for this change is the impact: E164 starting with calling code without + sign is used in:
- Numeric Transformation
- Call ACLs
- Pricing Plans
- Routing Patterns
- The whole universe

Does it worth? Or not? Discussion is open.

Kaian:
This is a breaking change that requires:

- Adapt all CallACLPatterns
- Adapt all MatchListPatterns
- Adapt all RoutingPatterns
- Adapt all TargetPatterns
- Adapt all LcrRules

and so on....

I think, this should be in 2.0

PR #313, opened 2017-10-31
SIP Proxies E164 numeric transformations

Kaian:
This PR aims to implement #89 and #100 

The following commits contains two changes to achieve standarization of all numbers in the platform:

## Add a Country next to every Number input field 

This way the number is fully identified and unique worldwide. To make easier for administrators, the default value for the country will be the company assigned country so, most of the time no change is required.

The scheme change includes setting all existing fields in database with current Company's country, but some manual fixes may be required.

## Standarize numbers in Users SIP Proxy

The fields used for number transformations (Country/AreaCode/OutboundPrefix) for Users/Friends/RetailAccounts has been replaced by a single selectbox to select transformation rules. Transformation Rules used by Peering Contracts, Users and RetailAccounts have been merged into a single table that is prefilled with one generic transformation for each country. That prefilled rules are shared by all the existing brands in the platform, but new ones can be created by brand for specific required transformations. We have added a transformation rules generator to make easier the process of adding those new brand transformations.

## Remove all numeric transformations from Application Servers
Now all the platform will use E.164 numbers internally with plus sign. Transformation of this numbers into final presentation will be done in the proxy just before sending the information to the endpoint. This removes all the existing madness logics in pbx (that was not even able to implement multiple dial scenarios).


Kaian:
I have also added migration from existing RoutingPatterns, GenericCallACLPatterns and CallACLPatterns.

PR #331, opened 2017-11-14
webadmin: remove pending old model references

mmadariaga:
Reviewed klear filters


library/IvozProvider/Klear/Filter/QueueStrategies.php
```diff
@@ -1,18 +1,23 @@
 <?php
 
-use \IvozProvider\Mapper\Sql as Mapper;
+use Ivoz\Ast\Domain\Model\Queue\Queue;
```
Kaian: I think it should be better to work with `Ivoz\Provider\Domain\Model\Queue\Queue` here.

The other entity is used to preload asterisk data only.

The below $_queue class attribute is Provider QueueDTO too

PR #332, opened 2017-11-14
core: remove old mappers and models

mmadariaga:
Old models, mappers and rest controller are not needed anymore


Kaian:
This PR stat is `+0 −261,233`

We need more changes like this.

PR #339, opened 2017-11-20
recordings: refactored recordings encoder

Kaian:
Update recordings encoder to use new entities and repositories.

All limits and notifications are still pending, as may be replaced by a newer generic limit logic (see #343)


kamailio/recordings/templates/emailBrands_body.tmpl
```diff
@@ -1,28 +0,0 @@
-This is an auto-generated message.
```
mmadariaga: I would keep templates. What do you think about it?

Kaian: I'm ok with that, I will move them to `microservices/recordings/templates/` until we implement #342 someday.

Thanks for the review :)

Kaian:
Pending implementing #343 to remove all commented code.

PR #344, opened 2017-11-21
Boss assistant whitelist

### LINKED ISSUE #334: Switch boss/assistant to use matchlists
Related to #211 , this change seems useful: instead of having a quite unusable whitelist regexp, make exceptions using matchlist feature.

Migration should create a matchlist with current values of the regexp, so that the change is transparent.

Kaian:
This PR replaces current string field for boss-assistant exception regular expresion by a matchlist.

This matchlist works as a white list so origins that matches can call directly to the boss without passing through the assistant.

The scheme commit also converts existing regular expresions to MatchLists.

This PR fixes #334


Kaian:
Updated documentation with the changes in the portal

PR #348, opened 2017-11-23
Call acl rework using Match lists

### LINKED ISSUE #211: Call ACL Patterns vs Match Lists
You have added a new feature to create "Match Lists".

These lists are used on External call filters and don't know if somewhere else.

Couldn't those lists be used in "Call ACLs"? Could Match Lists replace "Call ACL patterns"?

Kaian:
This PR introduces the following changes:

- Add a new Brand section in portal for MatchLists
- Show Brand's MatchLists in its Companies MatchList section in read only
- Remove all GenericCallACLPatterns and CallACLPatterns
- Relate CallACLs with existing MatchLists

Existing data will be migrated automatically converting GenericCallACLPatterns into Brand's MatchLists with a single pattern, and CallACLPatterns into Company's MatchList with a single pattern. Existing CallACLs will be updated to use these MatchLists.

While Company MatchList section will also display Brand's MatchLists, for now we only allow selecting them in CallACLs and not in the rest of sections that use MatchLits (ExternalCallFilters, ConditionalRoutes, OutgoingDDIRules, and so on).

Fixes #211


Kaian:
Updated documentation with these changes

Issue #180, opened 2017-06-26
portals: Merge two IVR sections into a unique one


cruzccl:
There are 2 IVR types now:

- IVR common: "press the user extension or wait" type of IVRs.

- IVR custom:  "press 1 to talk to Sales, 2 for IT, etc" type of IVRs.

Evaluate merging these two sections into a unique one.

Kaian:
I will schedule this for 2.0.0

- Create new tables to merge IVRCommon, IVRCustom and IVRCustomEntries
  - IVRs
  - IVREntries
  - IVRRelExtensions

- Update all routeTypes that use IVRCommon and IVRCustom to ivr 
  - Extensions
  - DDIs
  - ConditionalRoutes
  - ConditionalRoutesConditions

- Add new field to IVRs to toggle support for dialing extensions 
  - When enabled, a multiselect will alow excluding some extensions

- Improve IVREntry regular expression detection
  - If entry is numeric, handle them as a normal select option
  - Otherwise handle them as regular expressions

PR #351, opened 2017-11-28
ivr: merge IVRCommon and IVRCustom into IVR

Kaian:
This PR implements #180 merging IVRCommon and IVRCustom tables and updating all required routes labels and foreign keys.

This also includes the migration of those IVRs to new table appending '(common)' or '(custom)' label.


Kaian:
Thanks for the review :)

PR #352, opened 2017-11-28
Sanitize values within entities

mmadariaga:
Move sanitizes from services to entities


Kaian:
Just a note: some entities here like IvrCommon and IvrCustom has been removed in #351, so maybe some conflicts arise depending on merge order.

PR #366, opened 2018-01-11
Implement rest api

mmadariaga:
Implemented REST API based on api-platform


library/CoreBundle/Resources/config/parameters.yml
```diff
@@ -1,35 +1,32 @@
...
-        subject: '[ERROR] XMLRPC'
+        servers:
+            - jobs.ivozprovider.local
+        client:
+            timeout: 5000
+        mail:
+            from: no-reply@irontec.com
+            to:
+                - vozip@irontec.com
+            subject: '[ERROR] XMLRPC'
```
Kaian: Is there a reason to use 4 spaces indentation here?

Also, I prefer to remove the _mail_ section, that should not be used (and in case it's used should not contain irontec addresses)

Kaian: Nevermind all the file uses 4 spaces 😅

PR #374, opened 2018-01-31
CGrateS: Kamailio and BillableCalls

cruzccl:
This PR does several things related to kamailio-cgrates-cdrs integration:

- Add default CGrateS engine configuration JSON, enabling required components only.

- Add a systemd timer that loads tariff plans from database every fixed time interval. This will be removed in upcoming versions.

- Add to KamTrunks logic all the routes to control the call with cgr-engine help: ask for permission on establishment, warn when call starts and when it ends.

- It also modifies kam_acc_cdrs table and converts it into kam_trunks_cdrs: it only has kamtrunks cdrs and all the fields related to billing have been removed.

- Billing is done now in tp_cdrs CGrateS table.

- Change web BillableCalls section so that is show kam_trunks_cdrs entries for outbound calls, looking the price in tp_cdrs table.


library/Ivoz/Kam/Domain/Model/TrunksCdr/TrunksCdr.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * @return string
+     */
+    public function getGhostDuration()
+    {
+        $tpCdr = $this->getTpCdr();
+
+        if ($tpCdr) {
```
mmadariaga: This condition should be removed or completed

library/CoreBundle/Resources/config/orm_target_entities.yml
```diff
@@ -221,6 +221,8 @@ doctrine:
         Ivoz\Cgr\Domain\Model\RatingPlan\RatingPlan
```
mmadariaga: Rememeber to add Cgr namespace into ivozprovider/library/CoreBundle/Resources/config/doctrine.yml
  ```
orm:
    mappings:
      Cgr: ...
```

File: library/Ivoz/Kam/Domain/Model/TrunksCdr/TrunksCdr.php [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+    {
+        return $this->id;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isBounced()
+    {
+        if ($this->getBounced() == 'yes') {
```
Kaian: `bounced` is now numeric 1 or 0

PR #400, opened 2018-02-21
Add enable/disable toggle to user call forward settings

### LINKED ISSUE #378: Enable/Disable user call forward setting
Add to both user portal and company portal a toggle to enable/disable defined user call forward.

This way, instead of creating/deleting a call forward that you use a lot, you can simply disable/enable it.

cruzccl:
- User call forward can now be enabled/disabled from both admin portal and user portal.
- Disabled call forward settings are ignored.
- You can have multiple colliding call forwards but only one of them can be enabled at a time.

This PR fixes #378.

### DIFF
--- a/userweb/app/modules/detours/detoursEditCtrl.js
+++ b/userweb/app/modules/detours/detoursEditCtrl.js
@@ -30,6 +30,9 @@ angular
             $scope.extensions = extensions.data;
             Restangular.all('call-forward-settings').get(detourId).then(function(detour) {
                 $scope.detour = detour.data;
+
+                detour.data.enabled = '' + detour.data.enabled;
+
                 $scope.formDisabled = false;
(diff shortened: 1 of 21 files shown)


userweb/app/modules/detours/detoursEditCtrl.js
```diff
@@ -30,6 +30,10 @@ angular
             $scope.extensions = extensions.data;
             Restangular.all('call-forward-settings').get(detourId).then(function(detour) {
                 $scope.detour = detour.data;
+
+                detour.data.enabled = '' + detour.data.enabled;
+                console.log(detour.data);
```
mmadariaga: This is my mistake, remove it please

cruzccl: Done, thanks

PR #413, opened 2018-03-08
core: reviewed fields

mmadariaga:
Minor schema and visual enhancement


library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/HuntGroup.HuntGroupAbstract.orm.yml
```diff
@@ -36,13 +36,6 @@ Ivoz\Provider\Domain\Model\HuntGroup\HuntGroupAbstract:
       options:
         unsigned: false
       column: ringAllTimeout
-    nextUserPosition:
-      type: smallint
-      nullable: true
-      options:
-        unsigned: true
-        default: '0'
-      column: nextUserPosition
```
Kaian: We should also remove [this reference](https://github.com/irontec/ivozprovider/blob/bleeding/web/admin/application/configs/klear/HuntGroupsList.yaml#L31)

mmadariaga: You're right, thanks

Issue #394, opened 2018-02-19
Limit the number of voice channels per brand


cruzccl:
Currently you can limit the number of external voice channels per Company (vPBX or RetailClient):

- Extension 1001 calls to his cell phone: 1 channel.
- External number calls to a company DDI and 1001 answers: 1 channel.
- External number calls to a company DDI and a call forward diverts to an external number: 2 channels.

This channels are currently counted in KamTrunks.

This issue aims to:

- Change this selector so that it limits voice channels (both external and internal).
- Internally in 1.7, there will be two independent counters: one in KamUsers and another one in KamTrunks. The first that proxy whose counter arrives to configured limit will prevent this company to establish more calls.
- In Artemis, though, both proxies will share a unique counter.

Apart from this change, this issue will allow setting this limit per brand:
- Brand limit will be independent to company one: the first limit that is reached will prevent the brand/company to establish more calls.
- If you set it to 0, no brand-limit will be applied.

It is the same approach as with recording storage: there will be no control in assigning more channels to your companies than the channels you have as a brand.

cruzccl:
![](https://github.trello.services/images/mini-trello-icon.png) [Limit the number of voice channels per brand](https://trello.com/c/y1fJAxvn/8-limit-the-number-of-voice-channels-per-brand)

PR #414, opened 2018-03-08
Limit the number of voice channels per brand

cruzccl:
This issue replicates the changes made via #398 and fixed by #401 to achieve the goals described in #394 .


web/admin/application/languages/es_ES/es_ES.po
```diff
@@ -1085,6 +1086,9 @@ msgstr "Prioridad"
 msgid "Matching type"
 msgstr "Tipo coincidencia"
 
+msgid "Max call"
+msgstr "Límite llamadas"
+
 msgid "Max calls"
 msgstr "Límite llamadas"
```
Kaian: I would suggest using the already existing text _Max calls_ instead of creating a new one.

PR #417, opened 2018-03-13
tests: integarate jenkins ci with docker tests

Kaian:
The goal of this PR provide a base image and scripts to test pull requests based on their tags using Irontec JenkinsCI at https://jenkins.irontec.com


Kaian:
All tests are now fixed except ivozprovider-testing-api, but I think we could start using the integration by creating a new PR that fixes them.

There are lots of things to improve in the jobs, but let's hope we'll get there eventually.

PR #428, opened 2018-03-20
Add support for prepaid/pseudoprepaid billing

### LINKED ISSUE #389: cgrates: Add support for prepaid billing
Since 2.0, CGrateS sets the price for all outgoing calls with postpaid logic: all calls are granted and are assumed to be paid afterwards through a invoice.

This issue aims to add pseudo-prepaid logic:

- A call establishment will be granted only if the corresponding account has credit to assume the cost.
- Every X seconds the credit of the account will be decreased and in case it becomes 0, the call will be hangup.

cruzccl:
This commit aims to implement prepaid/pseudoprepaid billing methods at vPBX/retail client level:

- **Pseudo-prepaid**: authorizes the call on establishment (and limits duration) and only debits balance when the call ends.

- **Prepaid**: debits balances periodically and hangs up calls when credit is over.

Changes introduced:

- **_Billing Method_** in company / retail client new/edit screen: choose between postpaid/prepaid/pseudoprepaid billing method (default: _postpaid_).

- Added logic in Kamailio to add _ReqType_ param in CGrateS evapi requests.

- Added a database field (Companies.balance) that will have the balance associated with each prepaid/pseudoprepaid vPBX/retail client.

Handled in upcoming PR:

- Added new section at brand Level: **Balances**.
  - It will list clients (vPBX/retail) with prepaid/pseudoprepaid billing methods.
  - It will be possible to add credit for each client.
  - You will see current credit for each client.

- New microservice that asks CGrateS for each client balance and sets it in Companies.balance.

When prepaid/pseudoprepaid clients end their credit they cannot make new calls until more credit is charged.

This PR fixes #389


Kaian:
Please rebase before merging!

PR #434, opened 2018-03-28
Add enable/disable toggle to user call forward settings

cruzccl:
Add to both user portal and company portal a toggle to enable/disable defined user call forward.

This way, instead of creating/deleting a call forward that you use a lot, you can simply disable/enable it.


Kaian:
Everything seems ok. I would only add a couple more tests in `CheckUniquenessSpec.php` to test scenarios with existing disabled Call Forward Settings

PR #440, opened 2018-04-05
Hide Timings selector in RatingPlans list

cruzccl:
Timings selector is currently useless as it is not possible to add new timings.

This PR just hides it and makes '*any' default timing_tag for all rating plans (*any keyword behaves equal to 'ALWAYS' timing we used prior to this PR).


library/Ivoz/Cgr/Domain/Service/TpRatingPlan/InheritRatingPlanTag.php
```diff
@@ -17,9 +17,12 @@ public function execute(TpRatingPlanInterface $entity)
...
         );
 
-        $entity->setTimingTag(
-            $entity->getTiming()->getTag()
-        );
+        $timing = $entity->getTiming();
+        if (!is_null($timing)) {
+            $entity->setTimingTag(
+                $entity->getTiming()->getTag()
+            );
```
mmadariaga: We should set this service logic into the entity itself

PR #455, opened 2018-04-18
Remove unnecessary extension lifecycle

mmadariaga:
Voicemail/UpdateByExtension service is not necessary anymore


cruzccl:
My bad, I forgot removing them in #411 :(

Thanks!

PR #457, opened 2018-04-19
Decrement balance

mmadariaga:
Allow balance decrement in admin portal


web/admin/application/controllers/KlearCustomIncrementBalanceController.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
 use Ivoz\Provider\Domain\Service\Company\IncrementBalance;
+use Ivoz\Provider\Domain\Service\Company\DecrementBalance;
+use \Ivoz\Provider\Domain\Service\Company\AbstractBalanceOperation;
```
Kaian: Is absolute _use_ required here?

PR #460, opened 2018-04-19
Implement basic wholesale client interfaces

Kaian:
This PR adds a new Brand feature to enable Wholesale clients creation on brands, allowing access to its configuration sections in the portal.

This does not include any logic related to those screens, so the entered data won't be used until proxies logics are updated.


cruzccl:
Invoice and PeeringContract IMHO ;)

PR #464, opened 2018-04-24
Notify updated rates data

mmadariaga:
Set a key on redis every time rating data is updated so that cgrates knows that it must be updated


cruzccl:
After digging in CGrateS API, we can avoid reloading all tariffplans on these situations:

**Company/Retail/Wholesale creation**:

```cgr-console 'account_set Tenant="b1" Account="c1"```

**Company/Retail/Wholesale deletion**:

```cgr-console 'ratingprofile_rem Tenant="b1" Subject="c1" Category="call"'```
```cgr-console 'account_remove Tenant="b1" Account="c1"'```

**New Client-RatingPlan relation**:

```cgr-console 'ratingprofile_set Tenant="b1" Direction="*out" Subject="c1" Category="call" RatingPlanActivations=[{"ActivationTime": "2000-01-01T00:00:00Z", "RatingPlanId": "b1rp29"}]'```

I give you cgr-console commands instead of full RPC request/response, but you can get with:

```ngrep -d any port 2012``` 


Client-RatingPlan relation edition and removal can't be done with current API methods, but we may implement them if needed.


Thanks!

Issue #462, opened 2018-04-22
Invoice & Billable calls Date/Timezone Issue


afernandez2000:
Hi,

In a current setup I already setup a company with my local TimeZone (GMT-5) When I create an Invoice and specify the dates for example: Mar/1/2018 to Mar/31/2018 in the invoice are appearing calls from Feb/28/2018 I think is an issue with the TimeZone converting.

Attached some images.

![image](https://user-images.githubusercontent.com/1214749/39099610-a2f8a1ba-464b-11e8-97dd-3b646ea7af1f.png)

![image](https://user-images.githubusercontent.com/1214749/39099632-f2b819a6-464b-11e8-8d04-921e5745bf54.png)

mmadariaga:
Hi there,

Which version of IvozProvider are you using? Oasis (1.7) or Artemis (2.1)?

mmadariaga:
By the way,

Is your brand administrator in GMT-5 as well? 

Calls are fetched based on company timezone but represented in logged in user time zone.

afernandez2000:
Hi the version is 1.7.0. And yes the brand administrator and the Brand already have America/New_York timezone

mmadariaga:
Hi,

Could you apply the changes found in https://github.com/irontec/ivozprovider/pull/465/files into /opt/irontec/ivozprovider/library/IvozProvider/Gearmand/Invoices/Generator.php, regenerate the invoice and confirm that it works please?

Thanks

mmadariaga:
Sorry, I forgot to mention that in order to apply those changes a service restart is required :disappointed: : 
`sudo systemctl restart supervisor.service` 

Could you restart it and try again please?

afernandez2000:
Hi,

Yes now it works. Thank you very much

Alain Fernandez

PR #465, opened 2018-04-24
bug: fixed invoice timezone conversion

mmadariaga:
This PR aims to fix the issue reported in https://github.com/irontec/ivozprovider/issues/462


Kaian:
This looks good to me :+1:

PR #483, opened 2018-05-06
doc: install from packages repositories

manfer:
To install bleeding version we have to use debian stretch 9 as base system and bleeding as ivozprovider package and tayler as klear package.

### DIFF
--- a/doc/sphinx/basics/installation/debian_install.rst
+++ b/doc/sphinx/basics/installation/debian_install.rst
@@ -4,10 +4,10 @@ Debian packages install
 
 IvozProvider is designed to be installed and updated using Debian packages.
 More exactly, the current release is ready to be installed on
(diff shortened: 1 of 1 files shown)


manfer:
I have realized this is not a good idea. Bleeding is meant to be the development branch of last version, currently artemis. So the installation instructions should be for artemis here, isn't it?

I can correct this pull request if it is so.

Kaian:
Yep as mentioned in the other PR, bleeding dos will become artemis ones, so this change should not be updated.

Thanks!!

Kaian:
What about updating this PR with #484 changes (you must push -f the PR related branch with the other commit) and close #484?

The changes in #484 should have bleeding as destination :)

Kaian:
Also, don't worry about the failing test. It's complaining mostly about line wrapping in the translations, so I will fix it later :)

PR #484, opened 2018-05-06
doc: install from packages repositories

manfer:
To install artemis version from debian packages we have to use artemis for the ivozprovider package and tayler for klear package.


manfer:
I have realized that probably these changes had to be pull requested to bleeding branch. Then they will be added to artemis if they are accepted in bleeding. Should I do it that way?

Kaian:
Yep, replace commit in #483 with this one and we close this PR without merging. It will reach artemis on next release! :)

PR #486, opened 2018-05-07
Simplified lifecycle service config

mmadariaga:
Auto configure domain services lifecycle event and priority


Kaian:
Just pedantic fixes, I will review again after squashing the commits :+1:

PR #490, opened 2018-05-08
Fix initial terminal passwords

manfer:
Now the initial passwords for the demo terminals are hardcoded. This aims to randomize those initial passwords.

### DIFF
--- a/scheme/initial.sql
+++ b/scheme/initial.sql
@@ -3091,7 +3091,7 @@ CREATE TABLE `Terminals` (
 
 LOCK TABLES `Terminals` WRITE;
 /*!40000 ALTER TABLE `Terminals` DISABLE KEYS */;
-INSERT INTO `Terminals` VALUES (1,1,'alice','127.0.0.1','all','alaw',NULL,'invite','AUfVkn498_',1,'',NULL),(2,1,'bob','127.0.0.1','all','alaw',NULL,'invite','bob',1,'',NULL);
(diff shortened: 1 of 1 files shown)


cruzccl:
Hi @manfer!

Thanks for the contribution! :)

My 2 cents about it:

- Password for initial terminals created with the installation were meant for making first tests with IvozProvider as easy as possible.

- They are insecure and if you edit them, as they do not fulfill security requirements, they cannot even be saved.

- They should be equal to those documented [here](https://irontec.github.io/ivozprovider/en/oasis/internal_calls/configure_sipuacs.html), but alice password was unintentionally changed in 9f0d4f530.

In my opinion, we have several choices here:

**Continue with the "make first tests as easy as possible" approach**

2 options:

1. Revert change so that alice terminal has the password documented [here](https://irontec.github.io/ivozprovider/en/oasis/internal_calls/configure_sipuacs.html).
2. Change bobs default password to fulfill security requirements and update documentation accordingly.

**New approach: "make initial data as secure as possible even though it makes initial tests harder**

Your PR goes in this direction but, in my opinion, has this _minor_ drawback:

- Passwords are random, so someone making first tests should look for the password for each terminal.

- As autogenerated password do not meet security requirements, you cannot save the terminal after watching the password.


I would go for the easiest approach here (option 1), as initial data is not meant to remain and should make initial tests as easy as possible.

Opinions? :)

Thanks @manfer!!

manfer:
Agree with you. Better the simple approach with same password as documented. I didn't realize about that.

So alice password is the one that has to be reverted back to value alice. Should I change the commit?

Kaian:
If we are not updating the docs accordingly, I also prefer the first approach of just leaving alice password as _alice_ for easy getting starting.

If you are willing to change the commit for restoring the password you have my approval here :+1:

Thanks!

Kaian:
Dont worry about the failing test here we're updating them :smile:

PR #504, opened 2018-05-28
portals: fix Routing Pattern Groups GUI

manfer:
Routing Pattern Groups GUI is broken. Not totally sure but I think this is the correct fix.

### DIFF
--- a/web/admin/application/configs/klear/model/RoutingPatternGroups.yaml
+++ b/web/admin/application/configs/klear/model/RoutingPatternGroups.yaml
@@ -41,8 +41,8 @@ production:
             fields:
               - name${lang}
               - description${lang}
-              - regExp
-            template: "%name${lang}% %description${lang}% (%regExp%)"
+              - prefix
+            template: "%name${lang}% %description${lang}% (%prefix%)"
           filterClass: IvozProvider_Klear_Filter_Brand
 staging:
   _extends: production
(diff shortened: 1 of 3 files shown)


web/admin/application/configs/klear/model/RoutingPatternGroups.yaml
```diff
@@ -41,8 +41,7 @@ production:
             fields:
               - name${lang}
               - description${lang}
-              - regExp
-            template: "%name${lang}% %description${lang}% (%regExp%)"
+            template: "%name${lang}% %description${lang}%"
```
mmadariaga: It looks like this regression was introduced on https://github.com/irontec/ivozprovider/commit/a902f4ab7f19e592b0085d48a6d9df0002309482
Should we just rename the field @cruzccl?

mmadariaga:
Thank you for your contribution :)

web/admin/application/configs/klear/model/RoutingPatternGroups.yaml
Kaian: I think the correct fix is Renaming the field like  https://github.com/irontec/ivozprovider/commit/3b407da21878b0d9058f774549f5f7d4775a04e5 does

manfer: Now I understand better which was the real problem. Commit changed. Thanks.

Issue #503, opened 2018-05-27
bleeding call forward error


manfer:
I'm testing bleeding standalone installation.

When I configure a call forward inconditional on a user it works fine when he receives a call from another user. But if the user is the target of a DDI and I make an incomming call to that DDI then the call fails.

The only debug message I have been able to gather is from asterisk in verbose mode:

```
ERROR[1688][C-0000000e]: utils.c:1441 ast_carefulwrite: write() returned error: Connection reset by peer
```

This comes just after agi funcions in CallForwardAction had set some values:

```
[b2a69ae2] [CallForwardAction.php:145] AGI SetVar("PJSIP/proxytrunks-00000015", "REDIRECTING(from-num,i) = +34XXXXXXXXX")
[b2a69ae2] [CallForwardAction.php:146] AGI SetVar("PJSIP/proxytrunks-00000015", "REDIRECTING(from-tag,i) = 13")
[b2a69ae2] [CallForwardAction.php:147] AGI SetVar("PJSIP/proxytrunks-00000015", "REDIRECTING(from-name) = John Doe")
```

manfer:
I have enabled agi debug

```
<PJSIP/proxytrunks-0000001b>AGI Rx << SET VARIABLE REDIRECTING(from-num,i) "+34XXXXXXXXX" CallForwardAction.php:145
       > [7974fe33] [CallForwardAction.php:145] AGI SetVar("PJSIP/proxytrunks-0000001b", "REDIRECTING(from-num,i) = +34XXXXXXXXX")
<PJSIP/proxytrunks-0000001b>AGI Tx >> 200 result=1
<PJSIP/proxytrunks-0000001b>AGI Rx << SET VARIABLE REDIRECTING(from-tag,i) "13" CallForwardAction.php:146
       > [7974fe33] [CallForwardAction.php:146] AGI SetVar("PJSIP/proxytrunks-0000001b", "REDIRECTING(from-tag,i) = 13")
<PJSIP/proxytrunks-0000001b>AGI Tx >> 200 result=1
<PJSIP/proxytrunks-0000001b>AGI Rx << SET VARIABLE REDIRECTING(from-name) "John Doe" CallForwardAction.php:147
       > [7974fe33] [CallForwardAction.php:147] AGI SetVar("PJSIP/proxytrunks-0000001b", "REDIRECTING(from-name) = John Doe")
<PJSIP/proxytrunks-0000001b>AGI Tx >> 200 result=1
[...]

PR #509, opened 2018-05-28
Fix crash during external user call fordwards

Kaian:
Reported in #503 

We use the ORIGIN and CALLER channel variables to store entities ids in the form Entity#id

The ORIGIN variable is mainly used to apply OutgoingDDIRules on its presentation, but when the call is external, the ORIGIN is empty and no conversion or check is required. This PR updates the code to handle this case.

It also contains a test to validate this scenario with the dataset updated with latest tables changes.


manfer:
Just to let you know. Tested and working.

Kaian:
Awesome, thanks @manfer !

PR #520, opened 2018-06-06
Peer server orm definition

mmadariaga:
Some integer database fields were mapped as boolean in the orm


library/DataFixtures/ORM/ProviderPeerServer.php
```diff
@@ -41,7 +41,7 @@ public function load(ObjectManager $manager)
         $item2->setHostname("127.0.0.2");
         $item2->setPort(5060);
         $item2->setUriScheme(2);
-        $item2->setTransport(true);
+        $item2->setTransport(1);
```
Kaian: Actually, having an entry with transport != 1 will allow testing this problem won't happen again

PR #521, opened 2018-06-06
Enable http2 protocol

mmadariaga:
Enable apache http2 module and set it as preferred protocol


debian/ivozprovider-profile-portal.postinst
```diff
@@ -35,6 +35,7 @@ function setup_apache_config()
     [ ! -L /etc/apache2/mods-enabled/php7.0.load ] && /usr/sbin/a2enmod php7.0
     [ ! -L /etc/apache2/mods-enabled/headers.load ] && /usr/sbin/a2enmod headers
     [ ! -L /etc/apache2/mods-enabled/rewrite.load ] && /usr/sbin/a2enmod rewrite
+    [ ! -L /etc/apache2/mods-enabled/headers.load ] && /usr/sbin/a2enmod http2
```
Kaian: You mean http2.load here?

mmadariaga: You're right. Thanks

PR #552, opened 2018-06-22
Automatic invoice number

mmadariaga:
Set invoice number automatically if an 'invoice number sequence' is selected


Kaian:
LGTM!

I use to name the first commit with _scheme_ tag but that's not mandatory and all tests passed so shipit!

PR #563, opened 2018-06-27
Add support for Timings to Rating plans

### LINKED ISSUE #543: Add Timing criteria to RatingPlans
Currently RatingPlans group several DestinationRates and assign them a weight to solve collisions.

This issue aims to add Timing criteria so that DestinationRate group only applies in a given time period (L-V 8:00 - 14:00, weekdays, etc.).

This is a CGrateS feature that currently uses special keyword ***always** to apply all time.

Kaian:
This PR closes #543 by adding timing fields in TpRatingPlan structure.

Depending on how those fields are filled, the default _*any_ timing will be used or a new one will be created with the configured data.

### DIFF
--- a/debian/systemd/cgrates-loader.service
+++ b/debian/systemd/cgrates-loader.service
@@ -3,4 +3,4 @@ Description=Run cgr-loader with database tariff plans
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/cgr-loader -from_stordb -validate -cleanup -safety_key tp_changed_at
+ExecStart=/usr/bin/cgr-loader -from_stordb -validate -cleanup -timezone UTC -safety_key tp_changed_at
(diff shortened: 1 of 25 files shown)


cruzccl:
[This logic](https://github.com/cgrates/cgrates/blob/2f7b77f3a14d2f3b4c4e3a06f9f9a4abf4301e0f/engine/ratingprofile.go#L154) involving destinationRate selection makes this feature cause segfault on this situation:

- DestinationRate A with unique prefix +34 and a rate. Underlying rate is destA.

- DestinationRate B with unique prefix +34 and a rate. **Underlying rate is destB, even though the prefix is equal to destB.**

- RatingPlan with both A and B with different Timings:
  - A with Time In 00:00 weight 1.
  - B with Time In 12:00 weight 2.

Given logic will select only B, the one with higher weight, instead of selecting both of them (A for calls before 12 AM, B for remaining calls) as we would desire.

The reason for this is the sentence in bold: different destination.tag for same prefixes makes CGRateS think that they are not complementary destinationRates.


**Solutions:**

- CGRates patch.

- Modify DestinationRates lifecycles to reuse already existing destinations instead of creating new one for the same prefix (unique tpid+prefix in DB).


We need to think about this.

cruzccl:
After talking with CGrateS team (see cgrates/cgrates#1153) making a rework to avoid multiple destinations for the same prefix seems to be the way to go.

Kaian:
Updated with requested changes and rebased on top of `destination-rates-rework`, so its blocked until #605 merge

PR #589, opened 2018-07-16
Retail clients revamped

### LINKED ISSUE #557: New client type: retail client
#193 liberated Retail client name, converting existing Retail Clients to Residential Clients.

This issue aims to create a new retail client type:

- Can be defined as "_Wholesale-alike client, with incoming calls_".

- Retail Clients have retail accounts.

- They always speak SIP through KamUsers.

- Their calls do not traverse Application Servers (KamUsers<->KamTrunks signalling):

  - No call forwards.
  - No voicemail.
  - No access to virtual fax system.

- SIP authentication is mandatory in every new dialog.

- SIP registration is optional (if direct visibility exists, SIP registration is not needed).

- They have DDIs that can be pointed to desired retail account.

- They can only use clients' DDIs as origin in their calls, otherwise default outgoing DDI will be used.

- All their calls get recorded.

- No access to virtual fax system, but T.38 capable endpoints can use a retail account.

- They can use routingTags.

- Transcoding support as wholesale clients (see #487).

Kaian:
This PR fixes #557 and concludes the re-implementation of new Retail Clients

This new Retail Clients does not use Application servers, so they have less features than the original one (now revamped as Residential Clients). They work more or less like wholesale clients with support for DDI associations and incoming calls. 

Commits below only have entity and portal changes, proxy logics are not implemented yet.


Kaian:
Scheme errors on this are being caused by #588 , so I'm merging that and rebasing this PR

PR #579, opened 2018-07-04
Automatic invoices

mmadariaga:
Added invoice scheduler so that we can generate invoices automatically (once a month, once a week, custom frequency).

PR #591, opened 2018-07-18
Invoice scheduler fixes

Kaian:
Small fixes related to invoice schedulers #579 

web/admin:

- Reorder columns in Invoice Scheduler list
- Set default frequency to months
- Allow empty email
- Remove optional unassigned InoviceNumbers and InvoiceTemplate

core:
- Fixed calculation of next execution for Month and Year frequencies


Kaian:
Fixed the problem while deleting multiple rows.

The filter that checks each company can only be in one scheduler was also filtering in list screen.

Issue #526, opened 2018-06-07
Wholesale: add support for multiple rating profiles


cruzccl:
Wholesale clients' calls are now billed using the rating profile assigned to them, just the same way as remaining client types.

This means that calls are billed equally, no matter the routing tags they use.

There should be some relationship between them:

- Assigning rating profiles to routing tags instead of to wholesale clients?

- Create billing tag concept that link those concepts (routing tags - rating profiles) in a more dynamic way?

:thinking:

cruzccl:
**Pending:**

- [ ] Update simulators to work with routing tags

cruzccl:
Simulator changes made in #599

PR #594, opened 2018-07-20
Add Routing tags to Rating profiles

Kaian:
This PR add an optional field Routing Tag to Rating profiles to implement  #526

This would only apply to clients with routing tags (Wholesale) and will remain hidden for the rest of the clients. The code must be updated when #589 is merged.

Additionally, this add a business table for TpRatingProfile, called RatingProfile. This way, the data of RatingProfile will be exposed to API, web interfaces and so on, while the data of TpRatingProfile will be kept with all the required data for CGRates.


Kaian:
Updated Retail rating profiles screens with routing tags.

PR #605, opened 2018-08-01
Destination rates rework

Kaian:
Before this PR, data was being imported into TpDestinationRates rows, that splitted its info into TpRates and TpDestinations, generating one row per imported line.

This PR implements Destinations as master table of prefixes for each brand, so they are reused while importing new CSV files.

 - New business/domain tables:
     * DestinationRateGroups: formerly known as DestinationRates, now serves as storage for imported CSVs and groups imported DestinationRates rows
     * DestinationRates: formely known as TpDestinationRates, now creates TpRates and TpDestination rates for each CSV row. This also creates new Destinations if imported prefix does not exists.

  - CGRates tables:
     * tp_destinations: filled during Destination lifecycles
     * tp_rates: filled during DestinationRates lifecycles
     * tp_destination_rates: filled during DestinationRates lifecycles

This rework is required in order to implement #563


microservices/workers/src/Worker/Rates.php
```diff
@@ -60,22 +59,22 @@ class Rates
...
+        DestinationRateGroupRepository $destinationRateGroupRepository,
         DtoAssembler $dtoAssembler,
         EntityPersisterInterface $entityPersister,
         Logger $logger,
         Client $redisClient
     ) {
         $this->em = $em;
-        $this->destinationRateRepository = $destinationRateRepository;
+        $this->destinationRateGroupRepository = $destinationRateGroupRepository;
         $this->dtoAssembler = $dtoAssembler;
```
mmadariaga: I would replace dtoAssembler and entityPersister by EntityTools here

web/admin/application/configs/klear/model/DestinationRates.yaml
```diff
@@ -3,74 +3,97 @@ production:
...
+        type: box
+        position: left
+        icon: help
+        text: _("Use point as decimal delimiter - e.g. 0.02")
+        label: _("Need help?")
+    cost:
+      title: _('Per minute rate')
       type: text
-      isMultilang: true
+      pattern: "^[0-9]{1,6}[.]{0,1}[0-9]*$"
```
mmadariaga: Propagate this to the setter please

web/admin/application/configs/klear/model/DestinationRates.yaml
```diff
@@ -3,74 +3,97 @@ production:
...
-      required: false 
-    name:
-      title: _('Name')
+      required: false
+      default: true
+    connectFee:
+      title: _('Connection fee')
       type: text
-      isMultilang: true
+      pattern: "^[0-9]{1,6}[.]{0,1}[0-9]*$"
```
mmadariaga: This should be replicated into the setter: `Assertion::regex($var, $pattern);`

web/admin/application/configs/klear/model/Destinations.yaml
```diff
@@ -0,0 +1,48 @@
...
+      title: _('Name')
+      type: text
+      isMultilang: true
+      trim: both
+      required: true
+      maxLength: 55
+      default: true
+    prefix:
+      title: _('Prefix')
+      pattern: "^\\+[0-9]+$"
```
mmadariaga: You could put this regexp into the setter as well

PR #620, opened 2018-08-14
Update AGI wrapper with readLocution and playbackLocution

Kaian:
This commit adds two new functions (_readLocution_ and _playbackLocution_) to agi wrapper for properly checking Locutions in dialplan.

Original functions (_read_ and _playback_) will only handle the locution string or full path to the filename of a Locution entity.

### DIFF
--- a/asterisk/agi/src/Agi/Wrapper.php
+++ b/asterisk/agi/src/Agi/Wrapper.php
@@ -152,36 +152,39 @@ class Wrapper
         return $this;
     }
 
-
     /**
      *
-     * @param LocutionInterface|null $file
+     * @param LocutionInterface|null $locution
      * @param string $options
      */
-    public function playback($file = null, $options = "")
+    public function playbackLocution(LocutionInterface $locution = null, $options = "")
     {
-        if (is_null($file) || empty($file)) {
+        if (empty($locution)) {
             return;
         }
 
-        if ($file instanceof LocutionInterface) {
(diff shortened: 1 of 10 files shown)


asterisk/agi/src/Agi/Wrapper.php
```diff
@@ -152,36 +152,39 @@ public function progress($file = "")
...
      *
-     * @param LocutionInterface|null $file
+     * @param LocutionInterface|null $locution
      * @param string $options
      */
-    public function playback($file = null, $options = "")
+    public function playbackLocution(LocutionInterface $locution = null, $options = "")
     {
-        if (is_null($file) || empty($file)) {
+        if (is_null($locution) || empty($locution)) {
```
mmadariaga: You can remove is_null() here, empty() will do the job

asterisk/agi/src/Agi/Wrapper.php
```diff
@@ -213,25 +220,20 @@ public function read(LocutionInterface $locution, $timeout = 0, $maxdigits = 0)
             ->locutionPathResolver
             ->getFilePath($locution);
 
-        if (!file_exists($file)) {
+        if (file_exists($file)) {
```
mmadariaga: Looks like you should keep "!" here

Kaian: Woops! Thx!

Issue #529, opened 2018-06-07
LCR call routing


cruzccl:
#528 defined cost as the amount of money the brand operator pays to carriers for placing outgoing calls.

This issue aims to add call cost as a routing criteria, so that the call is routed to the carrier that implies less cost for the brand operator.

We need to evaluate:

- Dynamic routing - static routing relationship: when to use one and when the other?
  - Per client?
  - Per routing tag?
  - First LCR, then static?
  - Can both be combined?

- How to compare the cost of calls taking into account different charge periods and interval start.

cruzccl:
Tasks:

- [ ] Make Kamailio ask CGRateS for LCR carriers when load_gws() loads a dummy gateway with id 0  and insert them instead of dummy gateway.

- [ ] Lifecycles and database migration to create one tp_destination entry per RoutingPatterns entry.

- [ ] Changes in OutgoingRouting table, section and lifecycles to allow creating dynamic rules (new table OutgoingRoutingRelCarriers, dummy gateway insertion, etc.)

- [ ] Convert "All companies" into "Clients without specific rules".

- [ ] Change client route subsection to show specific rules only. If none, show "Clients without specific rules".

PR #634, opened 2018-08-22
TpDestinations for RoutingPatterns LCRs

Kaian:
This PR adds a relation between RoutingPatterns and TpDestinations required to implement #529 

TpDestinations can now belong to a Destination or RoutingPattern. The CGRates chosen tags are `bXdstX` for the first ones and `bXlcrdstX` for the second ones, but this can change if you find a better tag naming :)


cruzccl:
Finally discarded as final approach does not need to specify a destination in tp_lcr_rules table, *any is enough as filter is done by Kamailio LCR module.

PR #635, opened 2018-08-23
kamtrunks: Insert CGRateS LCR carriers in gw_uri_avp

cruzccl:
load_gws() Kamailio LCR module's function loads gateways to gw_uri_avp.

This new logic parses the elements of this AVP and calls CGRateS LCR whenever a
GW with gw_id=0 is found.

CGRateS gateways are inserted in the place of that dummy gateway.

This way, static and dynamic gateways are combined.

This PR is related to #529


cruzccl:
PR discarded. Changes in KamTrunks have been introduced in #645 (commit 3c84127).

PR #652, opened 2018-09-03
Trunks LCR lifecycles reworked

Kaian:
This PR improves the relationship between Kamailio LCR (kam_trunks_lcr_*) tables and IvozProvider Business tables (OutgoingRouting, Carrier, RoutingPattern, ..).

Lifecycles in charge of keeping these tables syncd have been reworked to consider corner cases (like adding or removing routing patterns to groups).


library/Ivoz/Kam/Domain/Model/TrunksLcrRule/TrunksLcrRule.php
```diff
@@ -29,4 +30,30 @@ public function getId()
...
+        $company = $outgoingRouting->getCompany();
+
+        if (!is_null($company)) {
+            // Company specific rule
+            return sprintf(
+                '^b%dc%d$',
+                $outgoingRouting->getBrand()->getId(),
+                $company->getId()
+            );
+        } else {
```
mmadariaga: There is no need for this `else`, you could just return `all companies` value

Issue #654, opened 2018-09-04
CGRateS reloads for LCR routes


cruzccl:
CGRateS reloads related to new LCR routes need some fixes:

- [ ] **New carrier to existent LCR route**

Should add _tp_changed_at_ redis-key.

- [ ] **Delete LCR route**

Two options here, both of them need a CGRateS patch:

- Add _tp_changed_at_ redis-key (also needs a CGRateS patch to cleanup lcr_* entries).
- ratingprofile_rem + lcr_rem (via CGRateS patch).

- [ ] **Remove carrier from existent LCR route**

Should remove ratingProfiles using:

```
cgr-console 'ratingprofile_rem Tenant="b1" Subject="cr2" Category="lcr_profile33"'
```

- [ ] **Remove ratingPlan from carrier used in LCR routes**

It should add _tp_changed_at_ even if the carrier is not used in LCR routes (previous bug).

- [ ] **Remove carrier used in LCR routes**

Add in lifecycles introduced in #464 one command per LCR route including deleted carrier:

```
cgr-console 'ratingprofile_rem Tenant="b1" Subject="cr2" Category="lcr_profile33"'
```

- [ ] **Add ratingPlan to carrier used in LCR routes**

Currently _tp_changed_at_ is added. Although it is enough, it would be better to use ratingprofile_set command (add to the logic added in #464) instead of reloading all tariffplans.

PR #663, opened 2018-09-05
core: CGRateS reload from lifecycle services simplified

Kaian:
This PR implements #654 taking the most basic approach and setting the CGRateS reload key in redis whenever a row in the following tables is **created, removed or updated** totally ignoring if the cause is an update in an OutgoingRouting, Carrier, Company or whatever other entitiy that uses CGRateS:

- tp_account_actions
- tp_destinations
- tp_destination_rates
- tp_lcr_rules
- tp_rates
- tp_rating_plans
- tp_rating_profiles
- tp_timings

This commits remove a lot of code. Although it worked as expected, we should try to keep licecycles as simple as possible to make debugging and maintainance easier. In the end, [premature optimization is the root of all evil](http://wiki.c2.com/?PrematureOptimization)


Kaian:
I want to discuss some changes before merging this. I think we can replace the periodic timer with a async worker like we do with kamailio reload services 🤔

microservices/workers/src/Worker/Cgrates.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Worker;
+
+use GearmanJob;
+use Ivoz\Core\Infrastructure\Domain\Service\Cgrates\ReloadService;
+use Ivoz\Core\Infrastructure\Domain\Service\Gearman\Jobs\Cgrates as CgratesJob;
+use Mmoreram\GearmanBundle\Driver\Gearman;
+use Symfony\Bridge\Monolog\Logger;
```
mmadariaga: Psr\Log\LoggerInterface would be a better option (plus it's autowireable)

cruzccl: Much more autowireable, yes :+1:

Issue #707, opened 2018-10-01
web/admin: emulate client from client list fails


cruzccl:
Clients can be emulated in two different ways:

- Using the icon next to Client Configuration module.

- Accesing to client list (vPBX, retail, wholesale or residential) and using the rightest options in each row.

This second method seems to be broken.

PR #710, opened 2018-10-02
web/admin: fixed retail emulation from list screen

mmadariaga:
There was an error on retail client emulation from list screen

This PR aims to fix https://github.com/irontec/ivozprovider/issues/707


mmadariaga:
Of course not

Issue #697, opened 2018-09-25
reRateCalls: improve feedback


cruzccl:
When CGRateS API request fails in [RerateCallService.php](https://github.com/irontec/ivozprovider/blob/a96368740f9a5a0d12aea5e986b6b1c7dda0b7f6/library/Ivoz/Core/Infrastructure/Domain/Service/Cgrates/RerateCallService.php#L66), pop-up gets blocked.

This issue aims to improve the user feedback so that no browser inspector is needed to know what is happening.

PR #711, opened 2018-10-02
Improve rerate feedback on error

mmadariaga:
Log API error responses to syslog and show an error message on klear

This PR fixes https://github.com/irontec/ivozprovider/issues/697

### DIFF
--- a/web/admin/application/controllers/KlearCustomTarificatorController.php
+++ b/web/admin/application/controllers/KlearCustomTarificatorController.php
@@ -631,7 +631,33 @@ class KlearCustomTarificatorController extends Zend_Controller_Action
                 \Ivoz\Core\Infrastructure\Domain\Service\Cgrates\RerateCallService::class
             );
 
-            $rerateService->execute($pks);
+            try {
+                $rerateService->execute($pks);
+            } catch (\Exception $e) {
+
+                $close = $this->_helper->translate('Close');
+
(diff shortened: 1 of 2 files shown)


web/admin/application/controllers/KlearCustomTarificatorController.php
```diff
@@ -631,7 +631,31 @@ protected function retarificate($pks)
...
             );
 
-            $rerateService->execute($pks);
+            try {
+                $rerateService->execute($pks);
+            } catch (\Exception $e) {
+
+                $close = $this->_helper->translate('Close');
+
+                $showErrors = ($e instanceof \DomainException) || (APPLICATION_ENV === 'development');
```
Kaian: Can this be extracted from the setting in application.ini ?

mmadariaga: Yes we can, but...

This is how it's done in klear:
 ```
        $phpSettings = $this->getInvokeArg('bootstrap')->getOption("phpSettings");
        $showErrors = false;
        if (array_key_exists("display_errors", $phpSettings) && $phpSettings["display_errors"]) {
            $showErrors = true;
        }
        if (APPLICATION_ENV == 'development' || $showErrors) {
            $showErrors = true;
        }
``` 

I mean, it's not been handled properly anywhere

Issue #719, opened 2018-10-08
Agi - Dial application g option duplicate on hunt groups


manfer:
On HuntGroupCallAction the g option is adeed
https://github.com/irontec/ivozprovider/blob/cfdac6e97a0c2330a3f75803eac7515739cc11a7/asterisk/agi/src/Agi/Action/HuntGroupCallAction.php#L60-L61

but the default dialplan already adds it
https://github.com/irontec/ivozprovider/blob/cfdac6e97a0c2330a3f75803eac7515739cc11a7/asterisk/config/dialplan/default.conf#L68

Most probable it is ignored by dial application but maybe would be better to remove the option in one place or the other.

Kaian:
It seems the other dialplan lines include some forced options (ivr, world, etc),  so I think the easiest would be to remove it from the agi and leave it in the dialplan. 

Are you willing to PR this change?

Thanks @manfer!

PR #720, opened 2018-10-08
remove duplicate g option

manfer:
This aims to solve https://github.com/irontec/ivozprovider/issues/719


asterisk/agi/src/Agi/Action/HuntGroupCallAction.php
```diff
@@ -58,7 +58,8 @@ public function process()
         $timeout = array_shift($huntGroupTimeouts);
 
         // Configure Dial options
-        $options = "ig";
+        // g option already added on dialplan
```
cruzccl: I would remove this comment line :thinking: 

Thanks!

Issue #729, opened 2018-10-15
No billable calls


manfer:
I don't know if this is an already known issue and a work in progress or if there is something wrong on my test bleeding installation but none of the calls is listed in billable calls lists.

After making an external call when I run,
```
systemctl status cgrates
```
I can see this last line,
```
oct 15 16:43:58 iv2 cgr-engine[1179]: CGRateS <b31e407> [1179]: Rating plan not found for destination +34xxxxxxxxx and account: b2:cr2, subject: *out:b2:call:cr2
```

which is really strange because the call is not rejected. In the call simulator the called number works too. There is no doubt a rating plan for that destination exists for sure.

Kaian:
Hi @manfer 

Billable calls was changed in #617 and now there is a new table that merges kamailio and CGRateS data. This process is done by a bsckground task invoked by a systemd timer.

Ensure that timer is running by running _systemctl start ivozprovider-cdrs.timer_

This new table allows bussiness fields and api integrations ignoring if billing features are enabled. 

Regards!

manfer:
Hi @Kaian 

It was in fact not running. I understand that probably none of the updates starts that timer.

By the way, shouldn't it be added on debian/ivozprovider.postinst on enable_services function?

Thanks.

Kaian:
Yes, it should be there for sure! Are you willing to PR it?

Thanks for reporting!

Kaian:
Fixed in #733

PR #733, opened 2018-10-16
enable ivozprovider-cdrs timer

manfer:
Fix for https://github.com/irontec/ivozprovider/issues/729


Kaian:
Thanks for contributing! 

We use to add prefix to commits with a tag of whats being changed. More info [here](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/en/commits.md)


Would you mind adding _pkg: _ prefix to your commit?

Thanks!!

Issue #747, opened 2018-10-24
debian packaging rules file


manfer:
https://github.com/irontec/ivozprovider/blob/4e319bcfd35827aa1184db859264d75b690b4282/debian/rules#L63-L64

Is the second line correct or is it a typo?

```
	dh_systemd_enable -p ivozprovider-cdrs       --name=ivozprovider-cdrs               --no-enable
```

If it is a typo, then same in here:

https://github.com/irontec/ivozprovider/blob/4e319bcfd35827aa1184db859264d75b690b4282/debian/rules#L72-L73

Kaian:
Hi @manfer 

Nope, that configuration is correct.

What is specified in those lines is that those two unit files (ivozprovider-scheduler.service and ivozprovider-cdrs.service) must be in a package called ivozprovider-scheduler. Package names are described in debian/control file.

Although not too common, in this case our package have multiple systemd services and timers :)

Thanks for reporting!

manfer:
Hi @kaian,

Thanks for clarifying.

I'm trying to find why on bleeding packages update and if I run dpkg-reconfigure too I'm getting this error:
```
ERROR 1062 (23000) at line 1: Duplicate entry 'xxx.xxx.xxx.xxx' for key 'domain_unique'
```
where xxx.xxx.xxx.xxx is the IP where ivozprovider is installed. One IP standalone installation.

But I don't know debian packaging too much and it is being hard to find.

And when I look into database scheme -at least the initial scheme- domain_unique is a unique key of Companies table and there is no duplicate entry, no even any entry with xxx.xxx.xxx.xxx value in domain_users column. So I'm lost.

Kaian:
Mos probably here:

https://github.com/irontec/ivozprovider/blob/bleeding/debian/ivozprovider.postinst#L75

This assumes you only have one company with a "domain IP trick" (see docs for info), and that IP is the Users proxy address.

Theorically that queries should not match any line or one if you are doing "the trick".

manfer:
I have run into mysql console the query on this line,


https://github.com/irontec/ivozprovider/blob/4e319bcfd35827aa1184db859264d75b690b4282/debian/ivozprovider.postinst#L91


And is the one that generates the error. I would say that like is not correct because it does not just match an IP but any domain with three or more dots, like:

a.b.example.com
a.b.c.example.com
...

I would say the correct implementation is,

```
    # Update DemoCompany domain to USERS_ADDRESS
[...]

PR #749, opened 2018-10-24
pkg: update demo company domain fix

manfer:
To solve the problem explained in https://github.com/irontec/ivozprovider/issues/747


manfer:
I'm not sure if it is allowed to remove the demo company and create a new one with the IP trick. If that's an option I would have to remove the id = 1 condition of the first where clause.

manfer:
Just thinking shouldn't the same be done with DemoBrand

```
    /usr/bin/mysql --user=root ivozprovider -e "UPDATE Brands SET domain_users = '$USERS_ADDRESS' WHERE domain_users REGEXP '^([0-9]{1,3}\.){3}[0-9]{1,3}$'";
```

manfer:
I think it already works for demo brand. But much better on another PR if documentation have to be changed too.

Thanks.

PR #751, opened 2018-10-25
microservices: call recordings in bounced calls

### LINKED ISSUE #549: microservices: call recordings in bounced calls
If a client calls to a existing DDI platform, call is bounced.

If call must be recorded for both clients involved (caller and callee), only a call recording is generated.

This issue aims to add logic to the recordings task to detect this scenario and link recoding file to both clients.

Kaian:
Improve Recordings encoder to handle multiple Cdrs entries for each Call-Id (like in bounced calls). The core tagged commit is not necessary if #734 is merged before this.

This PR fixes #549

### DIFF
--- a/microservices/recordings/src/Recording/Encoder.php
+++ b/microservices/recordings/src/Recording/Encoder.php
@@ -5,7 +5,9 @@ namespace Recording;
 use Ivoz\Core\Domain\Service\EntityPersisterInterface;
 use Ivoz\Kam\Domain\Model\TrunksCdr\TrunksCdrInterface;
 use Ivoz\Kam\Domain\Model\TrunksCdr\TrunksCdrRepository;
+use Ivoz\Kam\Domain\Model\UsersCdr\UsersCdrInterface;
 use Ivoz\Kam\Domain\Model\UsersCdr\UsersCdrRepository;
+use Ivoz\Provider\Domain\Model\Ddi\DdiRepository;
 use Ivoz\Provider\Domain\Model\Recording\RecordingDto;
 use Symfony\Bridge\Monolog\Logger;
 use Symfony\Component\Process\Process;
@@ -21,6 +23,11 @@ class Encoder
      * @var UsersCdrRepository
      */
     protected $usersCdrRepository;
+
+    /**
+     * @var DdiRepository
+     */
+    protected $ddiRepository;
 
     /**
      * @var EntityPersisterInterface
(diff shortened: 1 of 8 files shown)


microservices/recordings/src/Recording/Encoder.php
```diff
@@ -109,117 +119,149 @@ public function processAction()
...
-                $kamAccCdr = $this->usersCdrRepository->findOneByCallid($callid);
-                if ($kamAccCdr) {
-                    $type = 'ondemand';
-                    if ($kamAccCdr->getXcallid()) {
-                        // If call second leg, callee is who activated the recording
-                        $recorder = $kamAccCdr->getCallee();
-                    } else {
+            // Look if the conversation with that id has ended
+            $kamAccCdrs = [];
+            $kamAccCdrs = array_merge($kamAccCdrs, $this->trunksCdrRepository->findBy(['callid' => $callid]));
```
mmadariaga: ¿Could you call findOneByCallid here? I'm trying to keep every query into it's own method in order to improve maintainability

microservices/recordings/src/Recording/Encoder.php
```diff
@@ -109,117 +119,149 @@ public function processAction()
...
+            if (empty($kamAccCdrs)) {
+                $kamAccCdrs = array_merge($kamAccCdrs, $this->usersCdrRepository->findBy(['callid' => $callid]));
```
mmadariaga: ¿Could you move this into the repository? Just like the one on trunksCdrRepository

microservices/recordings/src/Recording/Encoder.php
```diff
@@ -109,117 +119,149 @@ public function processAction()
...
+                    // Check this ddi has recording enabled
+                    $ddi = $this->ddiRepository->findOneByDdiE164($recorder);
+                    if (!$ddi) {
+                        $this->logger->error(
+                            sprintf("[Recordings][%s] Unable to find DDI for %s\n", $hashid, $recorder)
+                        );
+                        continue;
+                    }
+
+                    if (!in_array($ddi->getRecordCalls(), array('all', $direction))) {
```
mmadariaga: php's in_array is quite misleading ¿Could you set strict flag here? more info http://php.net/manual/es/function.in-array.php#61491

microservices/recordings/src/Recording/Encoder.php
```diff
@@ -109,117 +119,149 @@ public function processAction()
...
+                        $this->logger->info(
+                            sprintf("[Recordings][%s] %s has no %s recording enabled. Skipping.\n",
+                                $hashid, $ddi, $recorder)
+                        );
+                        continue;
                     }
                 } else {
-                    $stats['skipped']++;
-                    $this->logger->info(sprintf("[Recordings][%s] Call with id = %s has not yet finished!\n", $hashid, $callid));
+                    if ($kamAccCdr instanceof UsersCdrInterface) {
```
mmadariaga: You can merge this `if` with the `else` above if you want :)

PR #754, opened 2018-10-26
Unit test

mmadariaga:
Service refactoring and added some unit test

### DIFF
--- a/library/Ivoz/Provider/Domain/Service/CallCsvReport/CreateByScheduler.php
+++ b/library/Ivoz/Provider/Domain/Service/CallCsvReport/CreateByScheduler.php
@@ -26,7 +26,6 @@ class CreateByScheduler
         LoggerInterface $logger
     ) {
         $this->entityTools = $entityTools;
-        $this->restClient = $restClient;
         $this->logger = $logger;
     }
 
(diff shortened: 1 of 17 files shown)


library/Ivoz/Provider/Domain/Service/CallCsvReport/CreateByScheduler.php
```diff
@@ -45,11 +43,12 @@ public function execute(CallCsvSchedulerInterface $scheduler)
         } catch (\Exception $e) {
             $name = $scheduler->getName();
             $this->logger->error(
-                "Call CSV scheduler #${$name} has failed: "
+                "Call CSV scheduler #${name} has failed: "
                 . $e->getMessage()
             );
 
             if ($report && $report->getId()) {
+                // Does this make sense?
```
Kaian: I image this service wants to update the last execution time no matter if it success or fails. This way the scheduler won't stop on first failure (correct me if I'm wrong but next schedule is calculated from last execution).

Does _finally_ works even if you throw the exception in _catch_?

Don't forget to remove that comment :stuck_out_tongue:

Kaian:
Rebase and merge this! :+1:

Issue #758, opened 2018-11-02
kamctlrc database name


manfer:
Not sure what is this file for. But shouldn't these lines

https://github.com/irontec/ivozprovider/blob/4554ad8a2685f21da546b07f1e584d474656fe5f/kamailio/trunks/config/kamctlrc#L25-L26

be

```
## database name (for ORACLE this is TNS name)
DBNAME=ivozprovider
```

manfer:
I suppose this can be totally ignored as kamctl is disabled in favor of systemctl

https://github.com/irontec/ivozprovider/blob/4554ad8a2685f21da546b07f1e584d474656fe5f/profiles/proxy/root/.bashrc.d/500-proxy-system#L14-L15

Kaian:
That's what I was also thinking,  but I leave it to @cruzccl

If the file ia not used and not required, it should be deleted IMO

cruzccl:
Yes, it is not used and could be deleted.

kamctl is not used as kamcmd is used (systemctl to start/stop service).

Regards,

PR #763, opened 2018-11-07
Remove kamctl config

manfer:
In reference to https://github.com/irontec/ivozprovider/issues/758

I don't think anything else is needed as the whole config directories are copied to /etc/kamailio on install. So probably there is not any specific reference to those kamctl config files anywhere unless I miss something.


cruzccl:
Thanks @manfer

Issue #13, opened 2017-02-27
GL156 Add support for anonymous calls


cruzccl:

Kaian:
This could be merged with #6 adding a company defaultOutgoingDDI and changing both User/Company/Friend OutgoingDDI to interpret Unassigned as Anonymous.

This will obscure errors when both Company and Users have no outgoingDDI assigned. A good way to solve this could be a Company or Brand setting that will _allow Anonymous calls_.

manfer:
I have tried to make a call without DDI but hasn't been possible.

I'm going to try to look into this. Is it AGI the one that stops a call attempt if there is no OutgoingDDI?

manfer:
> I'm going to try to look into this. Is it AGI the one that stops a call attempt if there is no OutgoingDDI?

I found it and I did a test with a retail and I had been able to make anonymous call through a provider that allows me to make those calls.

I'm going to prepare a pull request with an implementation for you to take into consideration.

Guigui37:
Any news regarding this ?

Issue #489, opened 2018-05-08
Wholesale: route by call-prefix


cruzccl:
It is usual that wholesale client prepend their call attempts with a prefix to modify somehow the way that calls is managed:

- Use a premium carrier (with minimum ASR/ACD).
- Bill calls differently (e.g. with different currency).

This issue aims to add this feature to wholesale clients:

- Currently whosale calls are routed by country-prefix and by client (just the same as non-wholesale clients).

- From now on, wholesale clients will have some non-colliding prefixes to modify the routing logic.

- Detected prefix will be an additional criteria in Outgoing Routing section.

cruzccl:
Fixed in #501

PR #767, opened 2018-11-11
Oasis anonymous call

manfer:
I had not the chance to test it yet properly but It would be great if you can have a look to this implementation of https://github.com/irontec/ivozprovider/issues/13 using a prefix. Just to know if it is fine to do it this way or if it is a bad idea.


Kaian:
Hi @manfer 

Thanks for the pull request!  We will valorate and discuss this internally, because right now we are up-porting all oasis features to artemis and we are not sure how this will be ported.

In artemis, all application servers (asterisk, agis, ...) only handle numbers in E.164 with plus sign, so old outboundprefix logics has been moved to kamailio@users using numeric transformations, that are in charge of transforming what user dialed to that standard format, but are not ready for optional prefixes. The only optional prefix users can choose is ruled by _Routing Tags_ feature (#489) that could be used to choose the desired outgoingDDI used (or even none of them for an anonymous call). Those tags are not implemented for vPBX so porting this feature involves quite some analysis and development.

To sum up, this seems totally ok to me, but will hold this a bit until we have a valid artemis implementation.

Thanks a lot!

cruzccl:
Hi @manfer,

Thank you for taking time to contribute a PR with a new feature, it is highly appreciated.

We have thought many times about developing a feature like this (in fact #13 is one of our oldest issues), maybe not prefix-enabled but something that allows making anonymous calls.

The reason why we haven't done it yet is that we have seen that different carriers provide anonymous calls in a different ways:

- Some need setting anonymous in user-part of From (or PAI) header as your PR does.

- Some others need an additional anonymous-like domain in such headers (e.g. anonymous.invalid).

- Some others need additional Privacy header.

- Etc.

We will eventually implement this in Artemis (as you know we are avoiding scheme changes in Oasis to avoid complicating migration to Artemis), but we need to read thoroughly related RFCs ([RFC 3323](https://tools.ietf.org/html/rfc3323), [RFC 3325](https://tools.ietf.org/html/rfc3325)) before to learn how to implement it correctly, so that it works with all carriers that follow given specifications.

That said, I'm afraid we cannot merge this PR in IvozProvider until we have done all this analysis. As always, you can talk to our commercial department to prioritize this feature in Artemis release.

Regards,

manfer:
Thanks for detail description of the problem.

Besides, I think my implementation lacks the ability to decide which route to use to make the call. If I'm not wrong, it is kamailio the one that decides which provider to use based on the configured outgoing routes. I was thinking of using the prefix on an outgoing route but I think kamailio will get the outgoing call number without the prefix so that outgoing routing including the prefix would be ignored.

cruzccl:
Yes, KamTrunks decides the outgoing carrier for each call and, as you say, the anonymous prefix could collide with numeric transformations and routing logic.

We close this issue then to focus in active ones.

Thank you and regards,

PR #768, opened 2018-11-12
Fix virtual fax system

cruzccl:
Fix several typos and make tests to ensure it works as expected.

### DIFF
--- a/asterisk/agi/src/Agi/Action/FaxReceiveStatusAction.php
+++ b/asterisk/agi/src/Agi/Action/FaxReceiveStatusAction.php
@@ -3,7 +3,7 @@ namespace Agi\Action;
 
 use Agi\Wrapper;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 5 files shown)


asterisk/agi/src/Agi/Action/FaxReceiveStatusAction.php
```diff
@@ -101,7 +102,8 @@ public function process()
             $this->agi->error("File $faxTIF does not exists.");
             // Mark this fax as error and save
             $faxInDto->setStatus('error');
-            $this->entityPersister->persistDto($faxInDto, $faxIn);
+            $this->entityTools->persistDto($faxInDto, $faxIn, true);
+            $faxInDto = $this->entityTools->entityToDto($faxIn);
```
mmadariaga: Maybe we should do this automatically within persistDto: Update both, faxInDto and faxIn models.

cruzccl: Will be done in a separate PR :)

PR #770, opened 2018-11-12
core: terminal friend uniqueness

### LINKED ISSUE #308: core: terminal/friend uniqueness
All devices that register to a sip domain must have unique authentication id (name attribute on friends, terminals and retail accounts).

Restricting that for retail accounts is simple because a retail client can only have retail accounts. It can't have terminals nor friends. So the database unique index on name/brand is enough for retail accounts.

But a company can have both terminals and friends. So it can't have two terminals with same name. Easy to handle, it is already a unique index on database (name/domain). It can't have friends with same name. Easy to handle, an unique index can be created on database (name/domain). But, besides, it can't have a terminal and a friend with same name. And this last is not easy to handle right now.

Don't know which is the best way to solve this.

Kaian:
This PR is like #753 for Terminals and Friends.

The added services check Terminal and Friends does't have the same name in the same Company (or, in other words, with the same domain).

The TerminalRepository already had a method to find by name and domain, but I have update its signature so now is like the Residential Device, Friend, Retail Account ones. That's the reason why an agi commit is present in the PR.

This fixes #308

### DIFF
--- a/library/Ivoz/Provider/Domain/Service/Friend/CheckUniqueness.php
+++ b/library/Ivoz/Provider/Domain/Service/Friend/CheckUniqueness.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Ivoz\Provider\Domain\Service\Friend;
+
+use Ivoz\Core\Application\Service\EntityTools;
+use Ivoz\Provider\Domain\Model\Friend\FriendInterface;
+use Ivoz\Provider\Domain\Model\Terminal\TerminalRepository;
+use Zend\EventManager\Exception\DomainException;
+
+/**
+ * Class CheckUniqueness
+ * @package Ivoz\Provider\Domain\Service\Friend
+ */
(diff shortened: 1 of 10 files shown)


library/Ivoz/Provider/Domain/Service/Friend/CheckUniqueness.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        return [
+            self::EVENT_PRE_PERSIST => self::PRE_PERSIST_PRIORITY
+        ];
+    }
+
+    /**
+     * Check username and domain is unique in the whole platform
+     *
+     * @param FriendInterface $terminal
```
mmadariaga: $friend

Kaian: Done!

PR #776, opened 2018-11-15
Update ivozprovider-profile-proxy.templates

powerpbx:
Appears to be a comma missing.  

See https://github.com/irontec/ivozprovider/issues/774

PR #781, opened 2018-11-19
Update ivozprovider-profile-proxy.templates

powerpbx:
Appears to be a comma missing.  So config step of profile-proxy install cannot be completed when locale is set to english.

See #776


Kaian:
Thanks @powerpbx !

PR #787, opened 2018-11-20
web/admin: Platform Default Notification Templates section

Kaian:
This PR adds a new portal section under Platform menu to allow administrators to edit default Notification templates. This new section is limited to edition, so new and delete have been removed.

The PR also includes the missing screens for Call CSV Notification Templates contents.


Kaian:
Pending documentation

PR #816, opened 2018-12-06
Add pplsip to sip scanner list kamailio.cfg

powerpbx:


cruzccl:
Added this change and same change in KamTrunks via a9e2542

Thanks @powerpbx :+1:

Issue #821, opened 2018-12-10
CSV Schedulers: wrong calculation of in and out dates


Kaian:
It seems that after some refactoring in CSV schedulers services, the _in_ and _out_ date range is calculated AFTER updating nextExecutionDate.

For daily reports this means that the generated file only contains current day calls, instead of the ones of the previous day.

PR #822, opened 2018-12-11
core: fixed CallCsvReport/CreateByScheduler service

mmadariaga:
Do not modify input argument before service execution, nextExecution was
being updated before report creation.

This PR aims to fix https://github.com/irontec/ivozprovider/issues/821

### DIFF
--- a/scheme/tests/Provider/Invoice/CreateBySchedulerServiceTest.php
+++ b/scheme/tests/Provider/Invoice/CreateBySchedulerServiceTest.php
@@ -1,0 +1,133 @@
+<?php
+
+namespace Tests\Provider\Invoice;
+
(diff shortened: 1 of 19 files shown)


scheme/tests/Provider/Invoice/CreateBySchedulerServiceTest.php
```diff
@@ -0,0 +1,133 @@
...
+        $callCsvSchedulerDto
+            ->setName('')
+            ->setUnit('week')
+            ->setFrequency(1)
+            ->setEmail('')
+            ->setNextExecution($nextExecution)
+            ->setCompanyId(2)
+            ->setBrandId(1);
+
+        /** @var InvoiceScheduler $callCsvScheduler */
```
Kaian: wrong scheduler class hinting

PR #807, opened 2018-11-29
Update TpRatingProfile.TpRatingProfileAbstract.orm.yml

powerpbx:
Fix `activation_time` data type to match cgrates schema.  
https://github.com/cgrates/cgrates/blob/48e10f9201ff1b2fe3047c8f3e5d78086b87fc1e/data/storage/mysql/create_tariffplan_tables.sql#L113

Helps prevents some strange things from happening.
https://github.com/irontec/ivozprovider/issues/801

PR #833, opened 2018-12-21
scheme: change rp.activation_time type

### LINKED ISSUE #801: No rating plan, artemis
If I go to Brand> Billing > Rating Plans and create a plan and simulate a call it returns a successful simulation.  However, I cannot get a successful simulation from Clients > Virtual PBXs(or on Retail).  I confirmed the Rating Plan was added to the Client.

`systemctl status cgrates` shows  

> Rating plan not found for destination +1NXXXXXXXXX and account: b3:c6, subject: *out:b3:call:c6
Where NXXXXXXXXX is a 10 digit north american phone number.

Maybe I am missing a configuration somewhere?  I have an all-in-one install.  I also upgrade to v2.8 available today but it did not make a difference.

I did find that `ivozprovider-cdrs.timer` was not running so I enabled and started that but it did not change anything.

cruzccl:
As reported by @powerpbx in https://github.com/irontec/ivozprovider/pull/807#issue-234711702, CGRateS does not handle properly rating profiles activation_time if this field has datetime type.

This PR changes type to varchar (as CGRateS original field) and needs https://github.com/irontec/cgrates/commit/72fa43d91251d413591665f20ed1b3239a9f59bf to work properly.

This issue closes #801 and #807.

Thanks @powerpbx  for your report.

### DIFF
--- a/library/DataFixtures/ORM/CgrTpRatingProfile.php
+++ b/library/DataFixtures/ORM/CgrTpRatingProfile.php
@@ -31,9 +31,7 @@ class CgrTpRatingProfile extends Fixture implements DependentFixtureInterface
             $this->setTenant('b1');
             $this->setCategory('call');
             $this->setSubject('c1');
-            $this->setActivationTime(
-                new \DateTime('2018-01-01 10:10:10')
-            );
+            $this->setActivationTime('2018-01-01 10:10:10');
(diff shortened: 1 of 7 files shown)


Kaian:
I would rename the test commit to something more related to the change. Something like _tests: update orm test with TpRatingProfiles changes_

PR #841, opened 2019-01-03
devops: Optimized composer autoloaders

mmadariaga:
Convert PSR-0/4 autoloading to classmap to get a faster autoloader.
Authoritative flag is discarded because it would not allow service customization by client


debian/rules
```diff
@@ -37,14 +34,13 @@ build_webuser: build_library build_webadmin
...
-	cd microservices/scheduler; composer --no-interaction --no-dev  --no-progress install
-
+	cd microservices/balances; composer --no-interaction --no-dev --no-progress --optimize-autoloader install
+	cd microservices/recordings; composer --no-interaction --no-dev --no-progress --optimize-autoloader install
+	cd microservices/workers; composer --no-interaction --no-dev --no-progress --optimize-autoloader install
+	cd microservices/scheduler; composer --no-interaction --no-dev --no-progress --optimize-autoloader install
 
 build_scheme: build_library
-	cd scheme ; composer --no-interaction --no-dev  --no-progress install
+	cd scheme ; composer --no-interaction --no-dev --no-progress --no-interaction install
```
Kaian: extra --no-interaction here, missing --optimize-autoloader

PR #838, opened 2018-12-27
Add support for calendar events

Kaian:
This PR add extra fields to Holiday dates to allow entering a start and end time and specific routing options.

Holiday Dates can now be a whole day event or just during a Time Period.
Optionally, it can override both locution and treatment. If this fields are left empty, the external call filter locutions and handlers will be used.

Fixes #22

PR #842, opened 2019-01-04
Calendar events tests

mmadariaga:
Added tests for features introduced in https://github.com/irontec/ivozprovider/pull/838

Includes some code refactoring in order to prevent errors

### DIFF
--- a/library/spec/Ivoz/Provider/Domain/Model/Calendar/CalendarSpec.php
+++ b/library/spec/Ivoz/Provider/Domain/Model/Calendar/CalendarSpec.php
@@ -1,0 +1,160 @@
+<?php
+
+namespace spec\Ivoz\Provider\Domain\Model\Calendar;
+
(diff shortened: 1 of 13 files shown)


library/spec/Ivoz/Provider/Domain/Model/Calendar/CalendarSpec.php
```diff
@@ -0,0 +1,118 @@
...
+            ],
+            true
+        );
+
+        $this
+            ->isHolidayDate(new \DateTime('2019-01-02'))
+            ->shouldReturn(true);
+
+        $this
+            ->isHolidayDate(new \DateTime('2019-01-02'))
```
Kaian: Duplicate test.

Instead of updating the date, you could test the same date in different timezone :)

mmadariaga: Done :)

PR #844, opened 2019-01-07
Entity datetime error prevention

mmadariaga:
Return cloned instances of entity datetime attributes in order to avoid side effects and invalid object state

### DIFF
--- a/scheme/src/EntityGeneratorBundle/Tools/AbstractEntityGenerator.php
+++ b/scheme/src/EntityGeneratorBundle/Tools/AbstractEntityGenerator.php
@@ -191,7 +191,7 @@ protected function __toArray()
  */
 public function <methodName>(<criteriaArgument>)
 {<criteriaGetter>
-<spaces>return $this-><fieldName><forcedArray>;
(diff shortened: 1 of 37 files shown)


scheme/src/EntityGeneratorBundle/Tools/AbstractEntityGenerator.php
```diff
@@ -1211,6 +1211,9 @@ protected function generateEntityStubMethod(ClassMetadataInfo $metadata, $type,
         $parentResponse = parent::generateEntityStubMethod($metadata, $type, $fieldName, $typeHint,  $defaultValue);
         $parentResponse = str_replace('(\\' . $metadata->namespace . '\\', '(', $parentResponse);
 
+        $prefix = '';
+        $subfix = '';
```
Kaian: _subfix_ means _suffix_? Or a child _fix_?

mmadariaga: suffix, thanks :)

PR #852, opened 2019-01-08
kamtrunks: allow outgoing calls if CGRateS is unreachable

cruzccl:
Before this PR calls where dropped if CGRateS connection was down.

This could lead to long duration service failure.

This PR:

- Skips CGRateS call grant if CGRateS is down.

- Treats GWs as externally rated (no CGRateS involved) when CGRateS is down.

After this PR, BillableCalls table will allow entries without price and cost
through non-externally-rated gateways.

Rerating logic will be modified too to allow retarification of these calls, implementing
offline billing using CGRateS CdrsV1.ProcessExternalCDR method.

### DIFF
--- a/kamailio/trunks/config/kamailio.cfg
+++ b/kamailio/trunks/config/kamailio.cfg
@@ -34,6 +34,19 @@
 # Maximum call duration: 3 hours
 #!define MAX_DIALOG_TIMEOUT 10800
 
+# Allow calls when billing engine is down
+#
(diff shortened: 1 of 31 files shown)


Kaian:
Well, not really, the statement is still valid as long as this is handled as an fatal error.

cruzccl:
> This is ok, although it goes against the statement: No plan, no call.

Finally this new behaviour is controlled though a Kamailio global variable that can be tunned without restarting KamTrunks. Read commit for further details.

PR #857, opened 2019-01-14
Update Xmlrpc.php

powerpbx:
This change allows all worker iterations on all reachable servers to complete.

Without the change, any unreachable servers will cause worker jobs to not completed in an unpredictable way. It may not do anything on any servers if the unreachable server is the first iteration, or it may get part way through before it hits the unreachable server and exits, or it may only miss the worker jobs on the last server if that is the unreachable server.

To reproduce, add multiple proxytrunks and/or proxyusers in ivozprovider.ProxyTrunks, ivozprovider.ProxyUsers tables and make one of those servers unreachable. Make changes/adds/removes to Global Config > Infrastructure > Application Servers which causes this worker to try reload dispatcher on all the kamailio servers.

The progress and error logging can be seen at `systemctl status supervisor`.

### DIFF
--- a/microservices/workers/src/Worker/Xmlrpc.php
+++ b/microservices/workers/src/Worker/Xmlrpc.php
@@ -156,7 +156,6 @@ class Xmlrpc
                 $this->logger->error(sprintf("[XMLRPC] Unable to send request %s to server %s [%s:%d]",
                     $method, $server->getName(), $server->getIp(), $port
                 ));
-                return false;
             }
         }
         return true;


Kaian:
The code mentioned is only executed for a proxy type so even for request for both proxy types this won't fail. 

We will remove the foreach loop here and make a repository method to get the single proxy information. 

Thanks for pointing us this!

Kaian:
This PR has been replaced by #860 by replacing the loop.

Thanks @powerpbx !!

PR #858, opened 2019-01-15
Update ivozprovider-profile-portal.postinst

powerpbx:
Fix a problem with gearman log file not being created.  You can see the error in gearman log when running `systemctl status gearman-job-server` after gearman is started.

`gearmand[7735]: /usr/sbin/gearmand: Could not open log file "/var/log/gearmand.log", from "/", switching to stderr. (Permission denied)`

### DIFF
--- a/debian/ivozprovider-profile-portal.postinst
+++ b/debian/ivozprovider-profile-portal.postinst
@@ -51,7 +51,7 @@ function setup_apache_config()
 
 function setup_gearman_job_server()
 {
-    sed -i -r 's/"(--listen=).*"/"\10.0.0.0"/' /etc/default/gearman-job-server
+    sed -i -r 's/"(--listen=).*"/"\10.0.0.0 --log-file=stderr"/' /etc/default/gearman-job-server
 }
 
 function setup_klear()


Kaian:
Hi @powerpbx 

We are trying to remove all /var/log files and make all programs use stdout, stderr and syslog. If no syslog daemon is installed (we usually remove the default installed one) all this logs are handled by journald. 

Journald logs could be redirected to a logs collector and it supports automatic log rotation, so its a better option than files and old logrotate.

AFAIK, the only logs currently in files not yet changed are the apache ones. 

Nevertheless, thanks for the contribution!!

debian/ivozprovider-profile-portal.postinst
```diff
@@ -51,7 +51,7 @@ function setup_apache_config()
 
 function setup_gearman_job_server()
 {
-    sed -i -r 's/"(--listen=).*"/"\10.0.0.0"/' /etc/default/gearman-job-server
+    sed -i -r 's/"(--listen=).*"/"\10.0.0.0 --log-file=stderr"/'
```
cruzccl: You forgot to add the file at the end of the line, it should be:
```
sed -i -r 's/"(--listen=).*"/"\10.0.0.0 --log-file=stderr"/' /etc/default/gearman-job-server
```

Thanks

powerpbx: Fixed

cruzccl:
Added to bleeding via 3d6f99d914fc75910f0796596646ec8486f82eac

Thanks for your time @powerpbx

Issue #856, opened 2019-01-14
xmlrpc worker only partially completes when at least one server unreachable


powerpbx:
This is the relevant section.

https://github.com/irontec/ivozprovider/blob/2b2bef350fdfbe6acf1db219f1168c357a5ca889/microservices/workers/src/Worker/Xmlrpc.php#L140-L162

Returning `return false` in the catch causes the worker jobs to not complete in an unpredictable way.  It may not do anything if the unreachable server is the first iteration, or it may get part way through before it hits the unreachable server and exits, or it may only miss the last one if that is the unreachable server. 

To reproduce, add multiple proxytrunks and/or proxyusers in ivozprovider.ProxyTrunks, ivozprovider.ProxyUsers tables and make one of those servers unreachable.  Make changes/adds/removes to Global Config > Infrastructure > Application Servers which causes this worker to try reload dispatcher on all the kamailio servers.

The progress and error logging can be seen at `systemctl status supervisor`.

Perhaps it would be better to not return on catch.
```
            } catch (\Exception $e) { 
                 $this->logger->error(sprintf("[XMLRPC] Unable to send request %s to server %s [%s:%d]", 
                     $method, $server->getName(), $server->getIp(), $port 
                 ));
            } 
        } 
        return true;
```
This way it will always complete all iterations and therefore all jobs on the reachable servers.

Issue #828, opened 2018-12-19
Add sentinel support from cgrates upstream


powerpbx:
Cgrates upstream now supports redis sentinel.

https://github.com/cgrates/cgrates/commit/25e3cda1d71e04cf35e0adab8b6ac324b3faf314

https://github.com/cgrates/cgrates/commit/83f21c17fe6138c32accba34e4cbad271ec98cbf

https://github.com/cgrates/cgrates/commit/34a5843beafc7b563f6090bad2a689f0a7fa5c3e

powerpbx:
What is the status of this?  Do you have any plans to rebase to current cgrates stable since there appear to have been a lot of changes since you forked it.

Kaian:
Hi @powerpbx 

Sentinel seems quite interesting and the way to go for proper redis high availability. We currently use pacemaker ms resource with data local address.
Sadly, picking those commits will require some time because they don't apply directly to our current fork.

Like for other softwares used in Ivozprovider (asterisk, kamailio, rtpengine,..) we choose a version and test it, adding required patches (sometimes from upstream) to make it as stable as possible. We stick to that version until we require a functionality from a newer version. Then the whole process of testing and patchings is repeated.

For CGRateS, as they don't follow a numeric versioning anymore, we've choosen a commit and won't be moving from there anytime soon. If you want us to schedule more resources to this, don't hesitate to conctact our sales departament at comercial@irontec.com. You can also submit a pull request with the feature and we will try to test it whenever we have some spare time. 

Otherwise this will be marked as future. 

Thanks

Kaian:
I will try to priorize this for next  releases. Any thoughts about your patched version with sentinel?

Did you stress tested the new version or tested with high amount of records?


Thanks for the info @powerpbx!

powerpbx:
We have not done stress testing of anything yet.  We are waiting for artemis to be production ready before we do that.

Kaian:
Included in 2.9.2

Thanks for all @powerpbx

Issue #649, opened 2018-09-03
Question: multi proxy


manfer:
On the scale graph of the readme, on the left it shows and image that recalls a standalone system (or a distributed system with just one instance of each service). On the middle shows an image that can be seen as an installation with two proxies (trunks and users) an multiple application servers. And on the right image an image that shows multiple proxies two.

But I always read here that the system is ment to use only two proxies so I can't imaging how the implementation of the right image could be done.

I have tried myself two solutions to try to handle proxy failover.

On a first solution I use two totally independent installations of ivozprovider, one to work as primary and the other as secondary. On voip devices I configure one line registered to the primary ivozprovider (for example client1.sip.ivozprovider1.test for this primary user proxy) and a second line registered with the secondary ivozprovider (for example client1.sip.ivozprovider2.test for this secondary user proxy).

PROS
- Very simple to understand and implement

CONS
- When something is added or reconfigured has to be done in both installs. Error-prone.
- CDRs are spread on two systems

These CONS could be solved by coding some software which makes use of an API to talk to both ivozprovider instances to automate configs and get CDRs.

On a second solution I make use again of two ivozprovider installations but they are not independent. Their Data Service are configured with a master-master replication (excluding just ProxyTrunks and ProxyUsers tables). I make use of SRV records for registration (for example client1.sip.ivozprovider.test for both user proxies) so while primary ivozprovider is running registrations goes to that instance and only if that ivozprovider instance fails then registrations goes to second instance.

PROS
- Things only need to be configured once
- No need to get CDRs from two places.

CONS
- Lack of simplicity.
- Posible unexpected errors arise that I don't take into account. Maybe there are more database tables that should be excluded of master-master replication. The scripts that generate CDRs should probably be run only in one of the instances (oasis).

I have no confidence on this second solution to try it on production.

For sure there are better implementations that I'm not trying. The hardest to understand for me is to make several proxies read/write to same data service without problems. Being the most worrying problem two proxies trying to write data at same time.

Am I complicating things too much? Any advice?

powerpbx:
I am presently thinking about the same things.  Ideally, at least two physically separate geo redundant installs with the databases connected master-master.

[...]

PR #860, opened 2019-01-15
workers: only send XMLRpc commands to proxy main address

Kaian:
Avoid looping while handling ProxyUsers and ProxyTrunks tables. Only one record is valid in the table.

This was discovered in #856 by @powerpbx and replaces #857 pull request


library/Ivoz/Provider/Domain/Model/ProxyUser/ProxyUserRepository.php
```diff
@@ -7,5 +7,9 @@
 
 interface ProxyUserRepository extends ObjectRepository, Selectable
 {
-
+    /**
+     * @inheritdoc
+     * @see ProxyUserRepository::getProxyMainAddress
```
mmadariaga: No need to self reference here :)

library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/ProxyTrunkDoctrineRepository.php
```diff
@@ -19,4 +20,25 @@ public function __construct(RegistryInterface $registry)
...
+
+        $qb
+            ->select('self')
+            ->addCriteria(
+                CriteriaHelper::fromArray([
+                    ['id', 'eq', 1],
+                ])
+            );
+
+        $result = $qb->getQuery()->getResult();
```
mmadariaga: ->getOneOrNullResult() could be interesting here

powerpbx:
A little confused why you are actively trying to block clustered features.  Perhaps we need to understand the direction of this project better.  We were under the impression it is supposed to be a distributed solution.

We have multiple geographically distributed proxies working now.  We use it as primary, secondary and not for load balancing.  Almost all hard SIP phones are designed for this purpose with primary and secondary server settings.

powerpbx:
Perhaps you can create another table(s) for this purpose.  The code works well as it is now so it would be a shame to lose this functionality.  It seems like a step backwards from the stated goal.

> From its beginning, IvozProvider was designed to be installed distributed between multiple machines...

cruzccl:
Hi @powerpbx,

My 2 cents about this thread:

- The only way to ensure that our IvozProvider development does not diverge from your fork and its modifications is to contribute them via PR so that they are included officially in the main project (after the PR gets the approval of IvozProvider development team). If not, we may introduce changes that require that you readapt your changes or, even worse, that make your changes un-adaptable.

- We are not tied to the changes made by any user using a modified version of IvozProvider: you can fork the project and do whatever you want (as long as you fulfill the license requirements described [here)](https://github.com/irontec/ivozprovider/blob/bleeding/LICENSE), but you cannot control where the original project goes.

- Note that the very same thing happens to us with our custom patches on the different softwares shipped with IvozProvider (Asterisk, Kamailio, rtpengine, CGRateS...): if our changes are not accepted or are not suitable for the original project, we have to readapt our patches every time we need changes introduced in the main project (e.g. CGRateS sentinel support in #828).

- This way to work is fair for the original project too, that gets enhancements from contributors that prefer contributing to the project instead of mantaining their patches (leaving aside the ethical fact of contributing to the people that shared their entire work on the first time).

All the above is the way I think about free software projects collaboration and I hope you agree with me.

Leaving all this aside, I have several technical doubts about your multi-proxy implementation and I will be pleased to talk about them in a new issue or in #649.

Regards and thank you for your thoughts and time,

powerpbx:
You should update your documentation if you don't plan on supporting multiple proxies.  Those images showing multiple proxy blocks is not representative of what this product is.

PR #886, opened 2019-02-06
ci: run ivozprovider-scheme test in mysql docker sidecar

Kaian:
Instead of using a testing machine for MySQL, run a sidecar docker so we can test all deltas in a volatile environment.


Kaian:
Thanks for the approval @mmadariaga 

Sadly this won't work flawless without https://issues.jenkins-ci.org/browse/JENKINS-46336 , so I will postpone this for a time.

PR #893, opened 2019-02-18
kamtrunks: fix DDI recognition logic

cruzccl:
This PR fixes DDI recognition current logic, that crashes if more than 100 DDIs are assigned to the same DDI Provider (loop exceeded).

New logic does:

- Gets all DDI Providers matching IP address and get their transformation sets.

- Apply transformation set to R-URI username.

- Find transformated DDI within brand's DDIs.


ironArt3mis:
Rebase completed.

PR #901, opened 2019-02-28
Enum constants

mmadariaga:
Make enums entity class constants


Kaian:
I love this <3

It would be possible to remove the word CHOICE from non-array constants? I think its more readable without that word.

Thanks!

mmadariaga:
Removing CHOICE from constant name would make impossible to have an enum with 'choices' value. Plus It reflects that the possible values are restricted and I don't really dislike it :)

Your call @cruzccl  ;)

Kaian:
Now you have no CHOICE but remove CHOICE from the constants. At least we agreed that CHOICE was not the best CHOICE for the variable identification tag and type was redundant for some fields names, so it wasn't a good CHOICE either.

😜

mmadariaga:
Having not a single CHOICE turned out to be a good CHOICE

ironArt3mis:
Rebase completed.

PR #905, opened 2019-03-06
Orm default number values

mmadariaga:
Numeric field default values should not be strings

### DIFF
--- a/library/Ivoz/Cgr/Domain/Model/TpDestinationRate/TpDestinationRateAbstract.php
+++ b/library/Ivoz/Cgr/Domain/Model/TpDestinationRate/TpDestinationRateAbstract.php
@@ -51,7 +51,7 @@ abstract class TpDestinationRateAbstract
      * column: max_cost
      * @var string
      */
-    protected $maxCost = '0.000';
+    protected $maxCost = 0.0;
 
     /**
      * column: max_cost_strategy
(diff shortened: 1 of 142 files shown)


ironArt3mis:
Rebase completed.

library/Ivoz/Cgr/Domain/Model/TpDestinationRate/TpDestinationRateAbstract.php
```diff
@@ -51,7 +51,7 @@ abstract class TpDestinationRateAbstract
      * column: max_cost
      * @var string
```
Kaian: This field is defined as decimal in orm files. Should be documented as float.

library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/Recording.RecordingAbstract.orm.yml
```diff
@@ -30,7 +30,7 @@ Ivoz\Provider\Domain\Model\Recording\RecordingAbstract:
       precision: 10
       scale: 3
       options:
-        default: '0.000'
+        default: 0.000
```
Kaian: Float fields have 0.000 as default, while decimal fields have 0.0
Is this intended?

Also precision and scale may not affect float fields.

mmadariaga: It turns out that you can specify float precision `duration float(10,3) NOT NULL DEFAULT '0.000'`. I'll set 0.0 as default value

PR #907, opened 2019-03-06
Strict change detection

mmadariaga:
Because a non strict comparation, some change were not being properly registered. Null to zero value change for instead.


scheme/tests/Provider/Carrier/CarrierLifeCycleTest.php
```diff
@@ -144,6 +144,10 @@ public function added_carrier_has_tpCdrStats()
         $changelog = $changelogEntries[0];
 
         $diff = $changelog->getData();
+        //Skip empty default values
```
Kaian: Missing space between `//` and comment :nerd_face:

PR #916, opened 2019-03-12
scheme: enable all features in DemoBrand

Kaian:
Enable all features in default brand DemoBrand, including Retail, Residential and Wholesale clients features.


Kaian:
This is not complete. Enabling Retail and Residential requires domain_users brand field to be filled with something (also updating Domains table). I must find a way to respect the field if it has a value and fill it and update Domains if it's empty

Kaian:
Updated with Domain queries

PR #918, opened 2019-03-13
core: update endpoints pickupgroups when user or extension changes

Kaian:
For vPBX users, PsEndpoints store some Terminal and User data. When user is assigned a new Terminal, the PsEndpoint data is updated with new User related data.

This PR fixes the field PsEndpoints.named_pickup_groups that was only being updated when the user is added or removed from a PickupGroup. If this user still doesn't have a Terminal assigned, [the service in charge or filling named_pickup_group](https://github.com/irontec/ivozprovider/blob/1c280aa7cf56eec8129f9c8503ac8ad97af5d5d3/library/Ivoz/Ast/Domain/Service/PsEndpoint/UpdateByPickUpRelUser.php) won't do anything at all.

This may break some tests 0:-)


ironArt3mis:
Rebase completed.

PR #919, opened 2019-03-14
Improved generators

mmadariaga:
- Enhancement: nullable relationship method/attribute documentation
- Bugfix: call sanitizeValues from entity trait instead of abstract entity

### DIFF
--- a/scheme/src/EntityGeneratorBundle/Tools/AbstractEntityGenerator.php
+++ b/scheme/src/EntityGeneratorBundle/Tools/AbstractEntityGenerator.php
@@ -133,7 +133,6 @@ public static function fromDto(
 <voContructor>
     $self = new static(<requiredFieldsGetters>);
 <fromDTO>
-    $self->sanitizeValues();
     $self->initChangelog();
 
     return $self;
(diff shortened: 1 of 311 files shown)


scheme/src/EntityGeneratorBundle/Tools/AbstractEntityGenerator.php
```diff
@@ -1177,15 +1177,56 @@ protected function generateEntityAssociationMappingProperties(ClassMetadataInfo
                 continue;
             }
 
-            $lines[] = $this->generateAssociationMappingPropertyDocBlock($associationMapping, $metadata);
+            $doc = $this->generateAssociationMappingPropertyDocBlock($associationMapping, $metadata);
+            $lines[] = $doc;
```
Kaian: Is there a reason to use a temporal variable `$doc` here?

mmadariaga: Removed

mmadariaga:
> squash commits

Squashed :)

PR #920, opened 2019-03-14
core: add timezone to every date time

mmadariaga:
Do not trust that php default timezone is set to UTC

### DIFF
--- a/library/Ivoz/Provider/Domain/Model/ExternalCallFilter/ExternalCallFilter.php
+++ b/library/Ivoz/Provider/Domain/Model/ExternalCallFilter/ExternalCallFilter.php
@@ -164,7 +164,7 @@ class ExternalCallFilter extends ExternalCallFilterAbstract implements ExternalC
         }
 
         $scheduleMatched = false;
-        $time = new \DateTime('now');
+        $time = new \DateTime('now', new \DateTimeZone('UTC'));
 
         /**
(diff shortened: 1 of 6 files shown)


library/Ivoz/Provider/Domain/Model/ExternalCallFilter/ExternalCallFilter.php
```diff
@@ -164,21 +164,21 @@ public function isOutOfSchedule()
...
+        $time = new \DateTime('now', new \DateTimeZone('UTC'));
 
         /**
          * @var ExternalCallFilterRelSchedule $externalCallFilterRelSchedule
          */
         foreach ($externalCallFilterRelSchedules as $externalCallFilterRelSchedule) {
             $schedule = $externalCallFilterRelSchedule->getSchedule();
             $company = $schedule->getCompany();
-            $timezones = $company->getDefaultTimezone();
+            $timezone = $company->getDefaultTimezone();
```
Kaian: I think this function is not properly coded originally.

Both `$company` and company `$timezone` could be obtained from current entity (and removed from the loop). Aside from that, and more related to this pull request, the variable `$time` should be in that company `$timezone` instead of UTC in order to check if _now_ (in company's timezone) is between the times defined in the schedule (also in company's timezone).

Kaian: Ok. Thanks for the clarification!

PR #926, opened 2019-03-20
core: do not rely on entity concrete implementation constants

mmadariaga:
Constants should be at EntityInterfaces, not at concrete implementations


Kaian:
Seems ok to, although I would approve this after conflict resolution

mmadariaga:
> Seems ok to, although I would approve this after conflict resolution
Unlocked :-)

PR #935, opened 2019-03-27
Agent data on commandlog

mmadariaga:
Set user data on Commandlog. Includes:
 - Ip
 - Admin id


web/admin/application/library/IvozProvider/Klear/Auth/User.php
```diff
@@ -150,4 +150,9 @@ public function getUniqueIdenForCache()
         // Devuelve un hash basado en las propiedades publicas
         return md5(print_r(call_user_func('get_object_vars', $this), true));
     }
+
+    public function __toString()
+    {
+        return 'Administrator#' . $this->id;
```
Kaian: I can't see `id` between this class attributes. :thinking: 
Maybe `_id` from `Klear_Model_User`?

web/admin/application/library/IvozProvider/Klear/Dynamic/Config/Base.php
```diff
@@ -69,8 +70,10 @@ public function init($siteConfig)
     {
         $this->_siteConfig = $siteConfig;
         if (\Zend_Auth::getInstance()) {
+            /** @var \IvozProvider\Klear\Auth\User $identity */
             $identity = \Zend_Auth::getInstance()->getIdentity();
             if ($identity) {
+                DataGateway::$user = $identity->__toString();
```
Kaian: what about `DataGateway::$user = (string) $identity`;

library/Ivoz/Provider/Domain/Model/Commandlog/CommandlogAbstract.php
```diff
@@ -307,6 +316,33 @@ public function getArguments()
...
+     * Set agent
+     *
+     * @param array $agent
+     *
+     * @return self
+     */
+    protected function setAgent($agent = null)
+    {
+        if (!is_null($agent)) {
+        }
```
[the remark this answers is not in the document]
mmadariaga: This is a generator error. I'll fix it into another PR

web/admin/application/library/IvozProvider/Klear/Auth/User.php
mmadariaga: Is attribute is injected in Klear\Auth\Adapter, so they coexist for some reason :(

I'll use _id anyway, it's a better option

PR #936, opened 2019-03-28
core: remove brand domain when retail and residential are disabled

Kaian:
This shouldn't happen often, as usually Brand features are not a toy, but for the sake of consistency, this sanitize removes brand domain when the field is hidden on the screen.


Kaian:
I prefer to wait and think how to this kind of attributes (based on features should be implemented).

Otherwise, this would require multiple Brand access to set its value (Create Brand -> Set Brand Features -> Update brand domain)...

PR #937, opened 2019-03-29
core: register commands on microservices

mmadariaga:
Avoid generic / fall back command info on database changes triggered by microservices


microservices/workers/src/Worker/Cgrates.php
```diff
@@ -59,6 +57,7 @@ public function reload(GearmanJob $serializedJob)
     {
         // Thanks Gearmand, you've done your job
         $serializedJob->sendComplete("DONE");
+        $this->registerCommand('Worker', 'cgrated');
```
Kaian: _cgrates_ here?

mmadariaga: Fixed. Thanks :)

Kaian: Good job!

Issue #943, opened 2019-04-01
Invoices not generated correctly


gil-obradors:
One invoice number sequence configured before the end of the month, like :
![image](https://user-images.githubusercontent.com/6016929/55330533-90552100-5491-11e9-83ca-4b75e8e94fc1.png)

One invoice scheduler before the end of the month:  ( client PBX ):
![Captura_2019-04-01_15-35-11](https://user-images.githubusercontent.com/6016929/55331580-e034e780-5493-11e9-9feb-36b678c20936.png)

With this log:
```
Apr  1 08:00:04 sip01 [6719]: changelog.INFO: EntityWasCreated > Ivoz\Provider\Domain\Model\Invoice\Invoice#13 > {"inDate":"2019-03-03 23:00:00","outDate":"2019-03-31 21:59:59","taxRate":21,"invoiceTemplateId":1,"brandId":1,"companyId":8,"numberSequenceId":4,"schedulerId
":1,"id":13} [] []
Apr  1 08:00:04 sip01 [6719]: changelog.INFO: EntityWasCreated > Ivoz\Provider\Domain\Model\FixedCostsRelInvoice\FixedCostsRelInvoice#9 > {"quantity":1,"fixedCostId":2,"invoiceId":13,"id":9} [] []
Apr  1 08:00:04 sip01 [6719]: changelog.INFO: EntityWasUpdated > Ivoz\Provider\Domain\Model\InvoiceScheduler\InvoiceScheduler#1 > {"lastExecution":"2019-04-01 06:00:04","lastExecutionError":"","nextExecution":"2019-05-01 06:00:00"} [] []
```

But the finish result:
![Captura_2019-04-01_15-22-17](https://user-images.githubusercontent.com/6016929/55331680-11151c80-5494-11e9-8e8d-ed1b33b1fac9.png)

No billable calls assigned, no invoce generated, and no invoice number sequence generated.
I have assigned manually one invoice number at this generated invoice, but not working.

If I enter to Client view, I can see 2000 registers with prices. I can simulate the cost of the calls.
If I enter to Brand view I can see all billable calls correctly listed.

If I can bring more logs? I have not find where...

Thanks a lot !

Kaian:
Hi @gil-obradors 

I haven't tested this bug yet, but out of curiosity... Are you able to manually create an Invoice for that company in that time period? 

Maybe it provides some kind of error or something. 

Thanks for all the testing!

gil-obradors:
[...]

PR #947, opened 2019-04-04
Invoice scheduler regression

mmadariaga:
Fixed automatic invoice number assignment through invoice scheduler.

This bug was reported at https://github.com/irontec/ivozprovider/issues/943 and it's a regression introduced in https://github.com/irontec/ivozprovider/pull/722


library/spec/Ivoz/Provider/Domain/Model/Invoice/InvoiceSpec.php
```diff
@@ -48,4 +50,94 @@ function it_is_initializable()
...
+            ->shouldBe(null);
+    }
+
+    /**
+     * Value must be null so that is not automatically regenerated
+     */
+    function it_sets_null_status_on_reset()
+    {
+        $this->dto->setStatus(InvoiceInterface::STATUS_CREATED);
+        $dto = clone $this->dto;
```
Kaian: The other tests clone *before* setting status to `$this->dto`

Is the order in this test intended?

mmadariaga: Yes, $this->dto is used for entity instatiation and local $dto for entity update. I wanted them both to have created status :)

ironArt3mis:
Rebase completed.

PR #953, opened 2019-04-08
core: do not require user priority on random strategy hunt groups

mmadariaga:
Fixed error when an user was added into random strategy hunt group


ironArt3mis:
Rebase completed.

PR #962, opened 2019-04-16
kam: replace xmlrpc with jsonrpcs

cruzccl:
Port and transport protocol will remain untouched:

- TCP 8000 for kamusers
- TCP 8001 for kamtrunks


library/Ivoz/Kam/Infrastructure/Kamailio/TrunksClient.php
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * @param array $payload
+     * @return int
+     */
+    private function getActiveCalls(array $payload)
+    {
+        $response = $this->sendRequest(
+            'dlg.profile_get_size',
```
Kaian: I would create a constant for this RPC method like all the other actions have

microservices/workers/src/Worker/Xmlrpc.php
```diff
@@ -5,6 +5,7 @@
 use Doctrine\ORM\EntityManagerInterface;
```
Kaian: This file is called _Xmlrpc.php_ but does not longer use XML at all, doesn't it?

mmadariaga: You're right. We'll rename them (jobs, etc.) in another PR

library/Ivoz/Kam/Infrastructure/Kamailio/TrunksClient.php
mmadariaga: I like it :)

PR #967, opened 2019-04-25
Active call endpoints

mmadariaga:
Expose platform/brand/company active calls through API

### DIFF
--- a/web/rest/brand/app/config/api/resources.yml
+++ b/web/rest/brand/app/config/api/resources.yml
@@ -50,3 +50,24 @@ Model\RatingPlanPrices:
             description: 'Rating plan group Id'
   attributes:
     pagination_client_enabled: true
+
+Model\ActiveCalls:
+  itemOperations:
+    get_my_active_calls:
+      access_control: '"ROLE_BRAND_ADMIN" in roles'
+      method: 'GET'
+      path: '/my/active_calls'
+      route_name: 'get_my_active_calls'
+      parameters: []
+      swagger_context:
+        produces:
+        - 'application/json'
+        tags:
+        - My
+        parameters:
+        - name: company
+          in: query
+          type: integer
+          required: false
+          description: 'Company Id'
+  collectionOperations: []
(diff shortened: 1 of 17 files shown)


web/rest/brand/src/Controller/My/ActiveCallsAction.php
```diff
@@ -0,0 +1,82 @@
...
+            return new ActiveCalls($activeCalls);
+        }
+
+        /** @var CompanyInterface | null $company */
+        $company = $this->companyRepository->find($companyId);
+        if (!$company) {
+            throw new NotFoundHttpException('Company not found');
+        }
+
+        if ($company->getBrand() !== $brand) {
```
Kaian: Is this better than comparing brand Ids?

web/rest/brand/app/config/api/resources.yml
```diff
@@ -50,3 +50,24 @@ Model\RatingPlanPrices:
             description: 'Rating plan group Id'
   attributes:
     pagination_client_enabled: true
+
+Model\ActiveCalls:
+  itemOperations:
+    get_my_active_calls:
+      access_control: '"ROLE_BRAND_ADMIN" in roles'
+      method: 'GET'
+      path: '/my/active_calls'
```
Kaian: Out of curiosity, will be _my_ the _namespace_ for all custom endpoints?  Or it's also possible to use _Provider_?

mmadariaga: There is no technical limitation about namespaces

web/rest/brand/src/Controller/My/ActiveCallsAction.php
mmadariaga: Doctrine uses the same instance for the same id, so the result would not change. It's is just shorter

Kaian: Oh, i've been comparing ids all my life thinking it was safer :(

ironArt3mis:
Rebase completed.

PR #981, opened 2019-05-07
Fixed 1 month interval scheduler operations

mmadariaga:
PHP does not apply add/sub intervals as we excepted (more info [here](https://stackoverflow.com/questions/3602405/php-datetimemodify-adding-and-subtracting-months)). 

Created a little helper so that 1 month schedulers work the way they were designed. Includes unit tests


library/Ivoz/Core/Domain/Model/Helper/DateTimeHelper.php
```diff
@@ -74,6 +74,86 @@ public static function ensureUTC(\DateTime $dateTime)
...
+            $response->modify('last day of previous month');
+        } else {
+            $response->modify('last day of this month');
+        }
+
+        return $response;
+    }
+
+    /**
+     * Adds an interval to a datetime and fixes month sub related issues
```
Kaian: _Subs an interval.._

ironArt3mis:
Rebase completed.

PR #987, opened 2019-05-14
Role based attribute filter on api

mmadariaga:
Expose filtered brand and company on lower level APIs


Kaian:
This may be a PITA but _Rol_ should be spelled _**Role**_ in english

PR #991, opened 2019-05-20
T38 passthrough fix

cruzccl:
After some T.38 passthrough real-life tests, we have learnt:
    
- t38_udptl_ec: 'redundancy' seems better than 'none'.
- t38_udptl_maxdatagram: 1440 seems an standard value (and fixes some problems).
- t38_udptl_nat: set to 'no' as rtpengine does relay t.38 traffic.

This last point is very important: no Application Server exposure is needed.


ironArt3mis:
Rebase completed.

PR #994, opened 2019-05-22
Fixed roles on refreshed token

mmadariaga:
Updated dependency for security fixes
Set custom user provider to refresh token so that new token roles are the correct ones (ROLE_USER currently)


ironArt3mis:
Rebase completed.

PR #995, opened 2019-05-22
tests: moved some pending orm tests from scheme to schema

mmadariaga:
We still have some orm tests in the old scheme directory


ironArt3mis:
Rebase completed.

PR #997, opened 2019-05-27
Fix rerating condition in findUnratedInGroup

cruzccl:
Replace wrong condition with the right one.

Fixes regression introduced in 310ce72d397dbdbb901d7f23c8cbe98bf5da94f5.


library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/BillableCallDoctrineRepository.php
```diff
@@ -75,7 +75,7 @@ public function findUnratedInGroup(array $pks)
 
         $conditions = [
             ['self.id', 'in', $pks],
-            ['self.invoice', 'in', $pks],
+            ['trunksCdr.cgrid', 'isNull'],
```
mmadariaga: We should add ['self.invoice', 'isNull', null] as well. There is no inner validation to avoid invoiced call rerating (just the one on web client).

PR #998, opened 2019-05-28
Retry to rate calls without cgrid on cdr to BillableCall migration

mmadariaga:
Try to avoid some manual work in case that cgrates was down during the call


ironArt3mis:
Rebase completed.

PR #1003, opened 2019-06-04
web/rest: allow file sizes up to 30M on destinationRateGroup importer

mmadariaga:
Requieres to increase upload_max_filesize and post_max_size on php.ini


ironArt3mis:
Rebase completed.

PR #1004, opened 2019-06-04
web/admin: make DDIProvider - DDI assignment optional

cruzccl:
It has no functional purposes and it is already nullable in database.

Make it clear that it can be left unassigned without any consequences.


doc/sphinx/locale/es/LC_MESSAGES/administration_portal.po
```diff
@@ -8276,3 +8276,11 @@ msgstr ""
...
 msgstr ""
 
+#~ msgid ""
+#~ "The :ref:`DDI Provider <DDI Providers>` "
+#~ "that provides this number. This relation"
+#~ " allow the platform to apply the "
+#~ "required :ref:`transformations` in order to"
+#~ " determine its standard form."
+#~ msgstr ""
+
```
Kaian: Remove unused translation

PR #1006, opened 2019-06-04
Allow unassigned route in Call Forward Settings

Kaian:
This allows creating Call Forward settings without handler.

Although this may not seem to have too much logic, it could be used to setup dial timeout by adding a NoAnswer CallForwardSettings to ResidentialDevices or Users


ironArt3mis:
Rebase completed.

PR #1013, opened 2019-06-11
Fix typo in transferred external calls.

glacroix:
I could not make call transfer to an external number. After some debugging, it turns out that the external number was not properly transformed so it was rejected by kamailio@trunks.

The bug is due to a tiny typo in kamailio.cfg of kamailio@users.


Kaian:
Thanks for your contribution!

PR #1018, opened 2019-06-17
Map nullable on delete cascade fk fields properly

mmadariaga:
Mark explicitly nullable on delete cascade foreign keys (default behaviour when nullablility is not defined) so that models and API are properly documented

### DIFF
--- a/library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/QueueMember.QueueMemberAbstract.orm.yml
+++ b/library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/QueueMember.QueueMemberAbstract.orm.yml
@@ -17,6 +17,7 @@ Ivoz\Provider\Domain\Model\QueueMember\QueueMemberAbstract:
         queueId:
           referencedColumnName: id
           onDelete: cascade
+          nullable: true
       orphanRemoval: false
     user:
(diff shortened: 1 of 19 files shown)


library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/QueueMember.QueueMemberAbstract.orm.yml
```diff
@@ -17,6 +17,7 @@ Ivoz\Provider\Domain\Model\QueueMember\QueueMemberAbstract:
         queueId:
           referencedColumnName: id
           onDelete: cascade
+          nullable: true
```
Kaian: This shouldn't be nullable

library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/QueueMember.QueueMemberAbstract.orm.yml
```diff
@@ -28,4 +29,5 @@ Ivoz\Provider\Domain\Model\QueueMember\QueueMemberAbstract:
         userId:
           referencedColumnName: id
           onDelete: cascade
+          nullable: true
```
Kaian: This shouldn't be nullable

PR #974, opened 2019-04-30
Unified service collections

mmadariaga:
- Use a single service collection instance for every binded event (We were using a instance per event: on commit, pre remove, etc).

- Unify how common and entity specific lifecycle services are registered.

- List binded services on service collections (Visual sugar, it's overridden on cache compilation and it will not show customer specific services).

PR #1025, opened 2019-06-20
Fixed rating plan duplicated weight error

mmadariaga:
Fixed custom message on RatingPlan weight collision 
Fixed fallback error handler invocation when custom error handler exists

### DIFF
--- a/library/Ivoz/Provider/Domain/Service/RatingPlan/PersistErrorHandler.php
+++ b/library/Ivoz/Provider/Domain/Service/RatingPlan/PersistErrorHandler.php
@@ -15,6 +15,7 @@ class PersistErrorHandler implements PersistErrorHandlerInterface
     const MYSQL_ERROR_DUPLICATE_ENTRY = 1062;
 
     const RATING_PLAN_DUPLICATED_WEIGHT = 40002;
+    const RATING_PLAN_DUPLICATED_WEIGHT_NAME = 'weight';
 
     public function __construct()
     {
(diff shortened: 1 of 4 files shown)


library/Ivoz/Provider/Domain/Service/RatingPlan/PersistErrorHandler.php
```diff
@@ -34,7 +35,7 @@ public function handle(\Exception $exception)
 
         $isDuplicatedWeightError =
             $pdoException->getErrorCode() === self::MYSQL_ERROR_DUPLICATE_ENTRY
-            && strpos($exception->getMessage(), self::RATING_PLAN_DUPLICATED_WEIGHT);
+            && stripos($exception->getMessage(), self::RATING_PLAN_DUPLICATED_WEIGHT_NAME);
```
Kaian: I would compare with === FALSE, because stripos can return 0 as a valid non-error value

library/spec/Ivoz/Core/Infrastructure/Domain/Service/Lifecycle/DoctrineEventSubscriberSpec.php
```diff
@@ -0,0 +1,153 @@
...
+        );
+
+        $commonHandlerCollection
+            ->execute(Argument::any())
+            ->shouldBeCalled()
+            ->willThrow(new \RuntimeException('runtime exception'));
+
+        $this
+            ->serviceContainer
+            ->get('lifecycle.common.error_handler')
```
Kaian: Can this be replaced by CommonLifecycleServiceCollection::class after #974 has been merged?

mmadariaga: It will break otherwise :)

Kaian: pity, i like constants :(

ironArt3mis:
Rebase completed.

PR #1029, opened 2019-06-24
web/admin: forced close of add/edit tabs on emulation

mmadariaga:
Close edit/add tabs when a brand/client is emulated so that you don't end up adding an entity with unexpected company/brand values


ironArt3mis:
Rebase completed.

PR #1037, opened 2019-06-27
Restrict file uploads on api

mmadariaga:
Filter download only files
Skip file download endpoints if no item getter is found


ironArt3mis:
Rebase completed.

ironArt3mis:
Rebase completed.

PR #1048, opened 2019-07-02
Db constraint review

mmadariaga:
Added ddiProviderRegistration unique constraint.
Set HolidayDates.voiceMailUser ON DELETE SET NULL.
Set custom unique DdiProviderRegistration Username-domain error message


ironArt3mis:
Rebase completed.

PR #1060, opened 2019-07-06
agis : avoid calls to a DND User members of a Queue

glacroix:
If a User is a member of a queue, but with DND activated, he should not get any call from this queue.

### DIFF
--- a/asterisk/agi/src/Agi/Action/QueueCallAction.php
+++ b/asterisk/agi/src/Agi/Action/QueueCallAction.php
@@ -42,7 +42,6 @@ class QueueCallAction
     {
         // Local variables to improve readability
         $queueMember = $this->queueMember;
-
         if (is_null($queueMember)) {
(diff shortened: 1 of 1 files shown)


Kaian:
Hi @glacroix 

Thanks for the contribution, looks ok to me. This requieres two members approval, so it won't be merged until next week.

Meanwile, you can rename the commit to add a prefix if you want and know how. We follow [these](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/en/commits.md) gidelines for commit tags, so in this case it would be `agis: avoid calls to a DND User members of a Queue`.

Thanks again for the contribution!

glacroix:
Thanks @Kaian, will try to stick to that in the future :)

Kaian:
No problem, I was talking about the commit text more than the Pull request text.

But don't worry :)

PR #1069, opened 2019-07-12
Initial Italian language strings for web/admin

dsmaldone:


Kaian:
Hi @dsmaldone 

This commits contains some MacOS files (.DS_Store) and its done against artemis branch instead of bleeding, but don't worry, it has the contents of it_IT.po file, so we take that part and start the process to add the language support.

We'll try to include this in 2.12 or next minor release. We're still currently working on 2.11.2.

Thanks!

Kaian:
This PR has been replaced with #1115

PR #1081, opened 2019-07-19
Avoid destination ambiguity in DDI-prefix and routing-tags

cruzccl:
This PR introduces restrictions in DDI-prefix and routing-tags format to avoid ambiguity in destination number. From now on:

- DDI-prefix must have this format: 1-3 digits ended by *.
- Routing-tag must have this format: 1-3 digits ended by #.

This way it is possible to distinguish prefix part from destination number.


Kaian:
Are you storing the prefix including the * or #?

PR #1106, opened 2019-08-21
Improved api filters

mmadariaga:
#### Description
Added is null / is not null filter on foreign keys
Do not filter by fields that are not in response model (Except for foreign keys)
Improved order by parameter documentation (enum instead of raw string)
Sort API filter parameters

#### Type Of Change 

#### Checklist:


ironArt3mis:
Rebase completed.

PR #1160, opened 2019-10-23
web/rest: added PUT /billable_calls/{callid}/rate endpoint on platofo…

mmadariaga:
…rm API

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:
- [ ] Commits are named and have tag following [commit rules](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/en/commits.md)
- [ ] Commits are split per component (schema, web/admin, kamusers, agis, ..)


web/rest/brand/src/Controller/Provider/PutBillableCallRatingAction.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if (count($calls) !== 1) {
+            $errorMsg = sprintf(
+                'Multiple calls found with callid %s',
+                $callid
+            );
+
+            throw new \DomainException($errorMsg, 500);
+        }
```
Kaian: If we leave it as is, it would be impossible to tarificate bounced calls within the same brand.
Maybe we should add extra conditions to avoid this scenario (maybe _direction_ with _outgoing_ as default value?)

mmadariaga: Updated repository query, thanks :)

ironArt3mis:
Rebase completed.

ironArt3mis:
Rebase completed.

PR #1194, opened 2019-11-22
core: fix terminal model template backup overwrite

Kaian:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
When a Terminal Model provisioning template is saved, the current template is renamed appending a .back to the file. If a file already exists with that name, an error raises because `rename()` is not able to overwrite files.

Following commit deletes the backup file before renaming current to .back, preventing the error.


ironArt3mis:
Rebase completed.

PR #1196, opened 2019-11-25
web/admin: expose extension on terminal provisioning

mmadariaga:
Cisco templates require terminal extension on specific provisioning step

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:


web/admin/application/controllers/KlearCustomRunCodeController.php
```diff
@@ -64,11 +65,21 @@ public function runSpecificCodeAction()
...
+                        'getExtension',
+                        []
+                    );
+
                     /**
                      * For backward compatibility reasons
                      * @deprecated this will be remove in ivozprovider 3.0
                      */
-                    $this->view->user->setLanguage($this->view->language);
+                    if ($extension) {
```
Kaian: We should add a data verification logic to ensure the terminal has a User and that user has SceenExtension assigned, throwing an exception otherwise.

That should make this checks unnecessary.

ironArt3mis:
Rebase completed.

PR #1197, opened 2019-11-27
Allow E.164 without plus as inbound number format

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

Calling in E.164 without + as national format caused transformations problems:

e.g. 34945945945 instead of 945945945 lead to +3434945945945

Current rules assumed that caller uses agreed formats, but some didn't. This new rules are more flexible.

#### Additional information

This PR changes both rules generators and predefined existing rules.


ironArt3mis:
Rebase completed.

PR #1240, opened 2020-01-13
schema: added index into tp_cdrs.answer_time

mmadariaga:
Add an index into tp_cdrs.answer_time so that it can be rotated

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


library/Ivoz/Cgr/Infrastructure/Persistence/Doctrine/Mapping/TpCdr.TpCdrAbstract.orm.yml
```diff
@@ -13,6 +13,9 @@ Ivoz\Cgr\Domain\Model\TpCdr\TpCdrAbstract:
     tpCdr_cgrid_idx:
       columns:
         - cgrid
+    tpCdr_createdAt_idx:
+      columns:
+      - created_at
```
cruzccl: I think tp_cdrs.answer_time would be a better option (it is equal to kam_trunks_cdrs.start_time and it does not change whenever call is billed).

mmadariaga: Changed :)

PR #1262, opened 2020-01-30
microservices: recover disableDestination logic

cruzccl:
Restores logic implemented in 3027797 and deleted in 0120f85

#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/microservices/workers/src/Worker/Rates.php
+++ b/microservices/workers/src/Worker/Rates.php
@@ -349,9 +349,14 @@ class Rates
             $this->logger->debug('About to insert DestinationRates');
             $tpDestinationRateChunks = array_chunk($destinationRates, self::CHUNK_SIZE);
             foreach ($tpDestinationRateChunks as $destinationRates) {
-                $this
+                $affectedRows = $this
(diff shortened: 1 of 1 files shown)


microservices/workers/src/Worker/Rates.php
```diff
@@ -349,9 +349,13 @@ public function import(GearmanJob $serializedJob)
...
             $tpDestinationRateChunks = array_chunk($destinationRates, self::CHUNK_SIZE);
             foreach ($tpDestinationRateChunks as $destinationRates) {
-                $this
+                $affectedRows = $this
                     ->destinationRateRepository
                     ->insertIgnoreFromArray($destinationRates);
+
+                if ($affectedRows > 0) {
+                    $disableDestinations = false;
+                }
```
Kaian: What this logic actually means?

Kaian: Oh I see, it seems to Reload CGRateS destinations only if new prefixes where added

PR #1400, opened 2020-05-28
kam: avoid empty pointer free warning log

cruzccl:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

Using these functions with vars:

- is_present_hf($var(header))
- remove_hf($var(header)

Causes warning messages like this:

WARNING: <core> [core/mem/q_malloc.c:476]: qm_free(): WARNING: free(0) called from core: core/action.c: do_action(1126)

This PR avoids using them to prevent those warning messages.

#### Additional information

There should be no logic change.

### DIFF
--- a/kamailio/trunks/config/kamailio.cfg
+++ b/kamailio/trunks/config/kamailio.cfg
@@ -977,75 +977,57 @@ route[SELECT_NEXT_GW] {
 
 route[PARSE_X_HEADERS] {
     # Extract routing tag
-    $var(header) = 'X-Info-RoutingTag';
-    route(PARSE_OPTIONAL_X_HEADER);
-    if ($var(header-value) != '') {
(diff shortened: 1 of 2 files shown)


kamailio/trunks/config/kamailio.cfg
```diff
@@ -978,71 +978,70 @@ route[SELECT_NEXT_GW] {
 route[PARSE_X_HEADERS] {
     # Extract routing tag
     $var(header) = 'X-Info-RoutingTag';
-    route(PARSE_OPTIONAL_X_HEADER);
+    $var(header-value) = $hdr($var(header));
```
Kaian, line 981: $var(header) is not used anymore, isnt it?

```$var(header-value) = $hdr('X-Info-RoutingTag');```

cruzccl: Evaluate second commit carefully (I will squash later if needed), check for errors please.

PR #1407, opened 2020-06-02
Phpstan level 4

mmadariaga:
Fixed some phpstan level 4 issues

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:


library/Ivoz/Provider/Domain/Service/RoutingPatternGroup/UpdateByRoutingPatternAndCountry.php
```diff
@@ -46,10 +46,7 @@ public function __construct(
     public function execute(RoutingPatternInterface $entity, CountryInterface $country)
     {
 
-        if (!$country->getZone()) {
-            /**
-             * @todo Every country has a value on db. Is this even needed?
-             */
+        if (!$country->getZone()->getEn()) {
```
Kaian, line 49: I don't think this condition is possible because Countries is a read-only table with fixed data.

mmadariaga: Removed

PR #1418, opened 2020-06-08
kamusers: Allow registers from terminals with no user-extension

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

_kam_users_ view excluded:

- Terminals linked to no user.

- Terminals linked to users with no screen extension.

This meant that these terminals won't fulfill SIP authentication (as _kam_users_ is used in AUTH route).

This PR stops excluding these terminals so that they can authenticate (with valid credentials, of course), but they cannot make any call.

#### Additional information

The reason behind this change is that deactivating users by unsetting terminal or extension could be useful in some situations, but could lead to REGISTER flooding (depending on SIP phone retrial timeout in case of failure). This got even worse if TCP was used for these registrations.

With this change, they will REGISTER (and even SUBSCRIBE and get NOTIFY-ed), but they won't be allowed to make (or receive) any call.


cruzccl:
#1419 prevents this flooding and prevents brute-force-attacks too.

We discard this PR: terminals with no user-extension will continue not being able to complete SIP auth.

PR #1423, opened 2020-06-10
Use return type hints

mmadariaga:
Updated generators to https://github.com/irontec/ivoz-dev-tools/tree/4.0.1 so that return type hints  are used (getters and php built in types only this far).

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


library/Ivoz/Provider/Domain/Model/Administrator/Administrator.php
```diff
@@ -36,7 +36,9 @@ public function getId()
      */
     public function setPass($pass = null)
     {
-        if ($pass === $this->getPass()) {
+        if ($this->isInitialized()
```
mmadariaga, line 39: getPass is null during instantiation and the getter requires a value now

library/Ivoz/Provider/Domain/Model/InvoiceNumberSequence/InvoiceNumberSequence.php
```diff
@@ -35,7 +35,7 @@ public function getId()
      *
      * @inheritdoc
      */
-    public function setVersion($version)
+    protected function setVersion($version)
```
mmadariaga, line 38: Setters are not public anymore

PR #1439, opened 2020-06-25
Transfomations e164 regression

Kaian:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
This PR is a partial regression from #1197 removing the Transformation Rules created in default Country Transformation Rule Sets and the Rules generator.

These rules allow dialing numbers in E.164 without plus format in Country rule sets that can lead to undesired behaviors if there are National numbers that start with the Country Code. An example of this is any Italian mobile starting with 39 that is also the Italy Country Code.

Before: Dialing 397000111 was transformed to +397000111 assuming the initial 39 was Italy CC
After: Dialing 397000111 will be transformed to +39397000111 without E.164 without plus rule 

Thanks @nitefood for reporting and debugging this.

### DIFF
--- a/library/Ivoz/Provider/Domain/Service/TransformationRule/GenerateInRules.php
+++ b/library/Ivoz/Provider/Domain/Service/TransformationRule/GenerateInRules.php
@@ -74,17 +74,6 @@ class GenerateInRules
         $ruleDto
             ->setTransformationRuleSetId($entity->getId())
             ->setType($type)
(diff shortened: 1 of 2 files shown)


Kaian:
Updated to only remove rules from default Country Transformation Rule Sets

ironArt3mis:
Rebase completed.

PR #1452, opened 2020-07-02
core: corrected bodyType parsing in CSV scheduler emails

nitefood:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Fixes a regression introduced by commit [9542131711a99b45f315d0bbd1687849d8afdd27](https://github.com/irontec/ivozprovider/commit/9542131711a99b45f315d0bbd1687849d8afdd27) that caused the Body Type to be ignored for Call CSV scheduled reports. HTML emails would go out with a 'text/plain' header that caused email clients to render it as plaintext instead of HTML (tested with Thunderbird v68.10.0).
Thanks @Kaian for the excellent support!


Kaian:
Thanks for your contribution!

PR #1453, opened 2020-07-03
pkg: fix storage permissions on fresh install

### LINKED ISSUE #1451: Logo Upload it's not working on artemis 2.15 at any Portal level
I tried at every level to Upload a logo for a brand/company portal, but klear is always returning "Undefined error"

Kaian:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:
- [X] Fixes an existing issue (Fixes #1451) 

#### Description
This is a regression from 2bf572708ab27dac3bb682ef3859f743330e9807 where storage permissions was changed to 777 after every upgrade. This causes some problems if the storage is distributed (DBRD, GlusterFS, and so on) and there are lots of files.

This PR revert the change only changing the root storage directory, so web interface can create required folders to upload images or logos.

#### Additional information

I still see some permissions problem in other cases. For example FaxFiles can be created from asterisk on incoming faxes and from the web interface for outgoing faxes. If the folders in storage are created during incoming faxes, outgoing ones that share storage path will fail to upload.

### DIFF
--- a/debian/ivozprovider-profile-common.postinst
+++ b/debian/ivozprovider-profile-common.postinst
@@ -11,4 +11,7 @@ if [ -z "$2" ]; then
     echo >> /root/.bashrc
 fi
 
+# Change storage perms
+chmod 777 /opt/irontec/ivozprovider/storage
+
 :


Kaian:
Looks like the FSO handling already creates files and directories with 777 so it can be read or updated by web interface or any other process, so what I mentioned in Addintional information is not a problem actually.

PR #1489, opened 2020-08-27
Billable calls endpoint name

mmadariaga:
Resolve BillableCalls.endpointId and set its string value into endpointName

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/BillableCall.BillableCallAbstract.orm.yml
```diff
@@ -87,6 +87,13 @@ Ivoz\Provider\Domain\Model\BillableCall\BillableCallAbstract:
         unsigned: true
         fixed: false
       column: endpointId
+    endpointName:
+      type: string
+      nullable: true
+      length: 255
```
Kaian, line 93: 255 chars looks like a lot to me. Not sure how MySQL handle the space of these varchars but the max lenght that the field will contain seams to be ResidentialDevice, RetailAccount or Friend name, which is 65 chars.

cruzccl (approved):

library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/Mapping/BillableCall.BillableCallAbstract.orm.yml
mmadariaga: Updated, varchars use length based size though :)
- CHAR column is fixed to the length
- VARCHAR columns are variable-length strings, values are stored as a 1-byte or 2-byte length prefix plus data

source: https://dev.mysql.com/doc/refman/8.0/en/char.html

PR #1510, opened 2020-10-22
kamusers: fix R-URI without dlg_vars

cruzccl:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:
- [x] Fixes an existing issue (Fixes #XXXX) 

#### Description

SAVE_CONTACT and FIX_RURI logic was broken under these circumstances:

- retail/wholesale calling to retail DDI.

- call was bounced to in trunks.

- dialog was handled twice in users.

- two dialog structures (one inbound, one outbound).

Accessing to dlg_vars in within-dialog requests always returns first dialog structure value (e.g. $dlg_var(direction) always outbound).

This causes R-URI being mangled incorrectly when callee hangs up (as saved contact from the other side was retrieved). This lead to BYE (and all within-dlg) misrouting.

This commit changes implementation so that no dlg_var is used.

### DIFF
--- a/kamailio/users/config/kamailio.cfg
+++ b/kamailio/users/config/kamailio.cfg
@@ -1733,9 +1733,8 @@ route[WITHINDLG] {
     route(ADAPT_REFERTO);
 
     # Fix overridden R-URI if needed
-    if (!$var(is_from_inside) && src_ip != myself && $dlg_var(contact) != $null && uri != $dlg_var(contact)) {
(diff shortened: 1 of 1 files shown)


Kaian:
When are the entries in this dialogs hash table removed?

cruzccl:
Here:

https://github.com/irontec/ivozprovider/blob/647f3b80066fbfb1fe39bdb25d84193a40ca69b9/kamailio/users/config/kamailio.cfg#L2499

Otherwise, they expire in 10800 seconds.

Kaian (approved):

PR #1553, opened 2021-02-04
kamusers: make multiContact per AoR configurable

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

This PR allows disabling parallel-forking (multiple SIP registrations using same SIP credentials) for Users, Friends, Residential Devices and Retail Accounts.

This way, only latest registered SIP device will ring, instead of ringing all registered devices simultaneously.


Kaian (approved):

Kaian (approved):

web/admin/application/configs/klear/model/Users.yaml
```diff
@@ -403,6 +403,29 @@ production:
...
+      title: _('Multi contact')
+      type: select
+      defautlValue: 1
+      source:
+        data: inline
+        values:
+          '0':
+            title: _("No")
+            visualFilter:
+              show: [""]
```
Kaian, line 416: FTR: Quotes are not required in empty lists, `[]` is perfectly fine.
Anyway, no need to change that in this PR.

PR #1587, opened 2021-04-21
Kam rpc redis jobs

mmadariaga:
Updated Kamailio RPC async calls. Moved from gearmand to redis queues

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


profiles/portal/etc/supervisor/conf.d/xmlrpc-daemon.conf
```diff
@@ -1,5 +1,5 @@
 [program:xmlrpc]
```
Kaian, line 1: Maybe we can take this opportunity to rename this services to kamrpc instead of xmlrpc (Filename and this header)

profiles/portal/etc/supervisor/conf.d/xmlrpc-delayed-daemon.conf
```diff
@@ -0,0 +1,8 @@
+[program:xmlrpc]
```
Kaian, line 1: Maybe we can take this opportunity to rename this services to kamrpc-delayed instead of xmlrpc (Filename and this header)
Also its better (if not required) if this header name is unique

web/admin/application/configs/klear/model/DdiProviderRegistrations.yaml
```diff
@@ -88,13 +88,13 @@ production:
...
-      source:
-        class: IvozProvider_Klear_Ghost_RegisterStatus
-        method: getDdiProviderStatusIcon
+#    statusIcon:
+#      title: _('Status')
+#      type: ghost
+#      dirty: true
+#      source:
+#        class: IvozProvider_Klear_Ghost_RegisterStatus
+#        method: getDdiProviderStatusIcon
```
Kaian, line 97: We don't need comments, we have git ! :smiley:

Kaian: Actually, what is this field disabled :thinking: ?

Kaian (changes requested):

Kaian (approved):
Awesome! :shipit:

PR #1608, opened 2021-06-15
[wip] New bouncer proxy

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

Bouncing is the internal name of outgoing external calls to DDIs handled in the platform (a user calling to its company's DDI, or calling to a DDI of another client in the platform).

These bounced calls are special in many ways:

- They traverse routing and billing logic as if they were being routed a carrier.
- In the last moment, calls are bounced from KamTrunks to KamTrunks.
- Despite being one SIP dialog, they are treated as 2 dialogs: one outgoing external and one incoming external.
- 2 CDRs will be created.
- They are indistinguible from a normal outgoing external call from the perspective of the caller and one incoming external call from the perspective of the callee.

This PR is a POC of rewriting current logic adding one additional piece of software: bouncer proxy.

This proxy will act as a B2BUA decoupling SIP dialog into 2 different dialogs making bouncing logic in KamTrunks unnecessary.

#### Additional information

With this new approach, relationship among brands in same platform may change, making possible that one specific brand sells PSTN connection to another brand in the same platform.

### DIFF
--- a/asterisk/agi/src/Dialplan/Trunks.php
+++ b/asterisk/agi/src/Dialplan/Trunks.php
@@ -66,8 +66,15 @@ class Trunks extends RouteHandlerAbstract
         /** @var \Ivoz\Provider\Domain\Model\Ddi\DdiRepository $ddiRepository */
         $ddiRepository = $this->em->getRepository(Ddi::class);
 
+        $brandId = $this->agi->getSIPHeader('X-Info-BrandId');
+        if (!$brandId) {
+            $this->agi->error("Missing X-Info-BrandId header, drop call");
+            $this->agi->decline();
+            return;
+        }
+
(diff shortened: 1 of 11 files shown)


adamgent:
Hi,

"making possible that one specific brand sells PSTN connection to another brand in the same platform."

Would this also make it possible for a "Brand A" to have "Brand B" as a wholesale/retail customer? e.g. for Brand B to use A as a Carrier/DDI Provider?

cruzccl:
Yes, that's the idea behind this change:

- brandA could have a retail client used by brandB as a carrier to place outbound calls using brandA carriers.
- the same retail client could be configured in brandB as DDIProvider too to receive calls using brandA ddiProviders.

PR #1613, opened 2021-07-09
Add retail to retail call-forward

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Retail account call-forward can only be routed to external numbers.

This PR makes it possible to call-forward to another retail account in the same retail client.

### DIFF
--- a/asterisk/agi/src/Dialplan/Retails.php
+++ b/asterisk/agi/src/Dialplan/Retails.php
@@ -3,12 +3,11 @@
 namespace Dialplan;
 
 use Agi\Action\ExternalNumberAction;
-use Agi\Action\ServiceAction;
+use Agi\Action\RetailCallAction;
 use Agi\Agents\RetailAgent;
 use Agi\ChannelInfo;
 use Agi\Wrapper;
 use Helpers\EndpointResolver;
-use Ivoz\Provider\Domain\Model\RetailAccount\RetailAccountInterface;
 use RouteHandlerAbstract;
 
 class Retails extends RouteHandlerAbstract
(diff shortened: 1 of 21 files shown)


asterisk/agi/src/Dialplan/Retails.php
```diff
@@ -98,9 +104,17 @@ public function process()
...
+            $dstRetailAccount = $this->endpointResolver->getRetailFromEndpoint($cfw_destination);
+            $this->retailCallAction
+                ->setRetailAccount($dstRetailAccount)
+                ->process();
+        } else {
+            // Remaining retail account calls are handled as external
+            $this->externalNumberAction
+                ->setDestination($exten)
+                ->process();
+        }
```
Kaian, line 118: Another possible approach here is passing CallForwardSettingId in a header and using CallForwardAction instead of these two actions. In that case, CallForwardAction would require some adaptations because it only supports Users/Residential forwards.


Example (require EntityManagerInterface and CallForwardAction in constructor) :

```
        $cfwSettingId = $this->agi->getSIPHeader('X-Info-Cfw-Id');

        /** @var CallForwardSettingsRepository $cfwRepository */
        $cfwRepository = $this->em->getRepository(CallForwardSetting::class);

        /** @var CallForwardSetting|null $cfwSetting */
        $cfwSetting = $cfwRepository->find($cfwSettingId);
 
        // Process Retail Call Forward 
        $this->callForwardAction
            ->setCallForward($cfwSetting)
            ->process();

File: asterisk/agi/src/Dialplan/Retails.php [hunk 2]
```diff
@@ -98,9 +104,17 @@ public function process()
         // Some feedback for asterisk cli
         $this->agi->notice("Processing outgoing call from \e[0;36m%s\e[0;93m to number %s", $retailAccount, $exten);
 
-        // All retail account calls are handled as external
-        $this->externalNumberAction
-            ->setDestination($exten)
-            ->process();
+        $cfw_destination= $this->agi->getSIPHeader('X-Info-Cfw-Destination');
```
Kaian, line 107: Avoid snake_case in variables, use $cfwDestination instead

mmadariaga (approved):

PR #1662, opened 2021-10-20
tests: shared home for Jenkinsfile tests

Kaian:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

    
This requires pre-creating a direcoty in Jenkins homedir/caches/ivozprovider
and aims to share downloaded dependencies for composer and npm between
builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,7 +23,6 @@ pipeline {
     // Environment configuration
     // ------------------------------------------------------------------------
     environment {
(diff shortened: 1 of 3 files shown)


Kaian:
Using a shared cache is a more cleaner approach than storing the cache inside the docker image and copy its contents to vendor before each run.

Sadly these changes are 1:30 min slower, so this is no an optimization. I leave the PR for future reference.

PR #1677, opened 2021-11-08
Psalm auto fix

mmadariaga:
Updated hints by running psalm --alter

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/library/Ivoz/Kam/Infrastructure/Kamailio/TrunksClient.php
+++ b/library/Ivoz/Kam/Infrastructure/Kamailio/TrunksClient.php
@@ -223,10 +223,7 @@ class TrunksClient implements TrunksClientInterface
         return $response->result;
     }
 
-    /**
-     * @return bool|int
-     */
-    public function isCgrEnabled()
+    public function isCgrEnabled(): bool
     {
         $response = $this->sendRequest(
             self::CGRATES_ENABLED_ACTION,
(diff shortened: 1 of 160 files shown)


ironArt3mis:
Rebase completed.

library/Ivoz/Kam/Infrastructure/Kamailio/TrunksClient.php
```diff
@@ -223,7 +223,7 @@ private function getActiveCalls(array $payload)
         return $response->result;
     }
 
-    public function isCgrEnabled()
+    public function isCgrEnabled(): bool|int
```
Kaian, line 226: Just a comment not meant to be handled in this pull request.

This returns non-boolean in case of error. I thing it would be more correct to raise an Exception in case of error instead of a multi-type return.

library/Ivoz/Provider/Domain/Service/BillableCall/UpdateByTpCdr.php
```diff
@@ -51,7 +51,7 @@ public function handle(DomainEventInterface $domainEvent)
         if ($isNotOutbound) {
             $msg = sprintf(
                 'Skipping %s call #%d',
-                $billableCall->getDirection(),
+                $billableCall->getDirection() ?? '',
```
Kaian, line 54: Just a comment not meant to be handled in this pull request.

Billable Call Direction should not be nullable. Inbound or Outbound. Other values should not be allowed.

Kaian (approved):

PR #1721, opened 2022-02-21
core: trim passwords

mmadariaga:
Trim passwords to avoid issues

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


cruzccl (approved):

library/Ivoz/Provider/Domain/Model/CarrierServer/CarrierServer.php
```diff
@@ -108,4 +108,13 @@ public function getName()
                 $this->getId()
             );
     }
+
+    protected function setAuthPassword($authPassword = null)
+    {
+        if ($authPassword) {
+            $authPassword = trim($authPassword);
+        }
```
Kaian, line 116: This check is not in the other setters that allow nullable arguments (like Terminals, RetailAccounts and so ?

PR #1756, opened 2022-04-04
Fix client portal filters

mmadariaga:
Both filters and forms use the same service in order to get select options. This was causing some isues in filters like showing only unassigned extensions. Check filterContext flag in order to avoid issues.

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:


Kaian (approved):
Good Code! Bad PR and branch names. :balance_scale: :shipit:

PR #1763, opened 2022-04-08
Fix Swagger gen path at install

adamgent:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:
- [ ] Commits are named and have tag following [commit rules](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/en/commits.md)
- [ ] Fixes an existing issue 

#### Description
Replace swagger export path web/swagger.json for Platform/Brand/Client REST API with correct path public/swagger.json

#### Additional information
The current export path for swagger.json is incorrect and so the task fails at post-install, resulting in no swagger API docs - also visible at https://irontec.github.io/

### DIFF
--- a/debian/ivozprovider-web-rest.postinst
+++ b/debian/ivozprovider-web-rest.postinst
@@ -14,7 +14,7 @@ setfacl  -R -m u:www-data:rwX -m u:root:rwX var
 # Create project cache
 bin/console cache:clear --no-warmup -q -n
 # generate swagger spec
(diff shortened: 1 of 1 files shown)


Kaian (approved):
Thanks for the contribution! :+1:

We have recently remove these commands in next major release branch in #1739, and we're planning to make its contents autogenerated in a future enhancement when anyone access that route.

cruzccl (approved):
Thanks a lot!

PR #1737, opened 2022-03-09
Enhanced rerating

mmadariaga:
Remove corrupted tp_cdr rows before rerating

#### Type Of Change 
- [X] Small bug fix
- [ ] New feature or enhancement
- [ ] Breaking change

#### Checklist:
- [X] Commits are named and have tag following [commit rules](https://github.com/irontec/ivozprovider/blob/bleeding/doc/dev/en/commits.md)
- [X] Commits are split per component (schema, web/admin, kamusers, agis, ..)
- [ ] Changes have been tested locally
- [ ] Fixes an existing issue (Fixes #XXXX)

PR #1807, opened 2022-06-01
Enhanced rerating

Kaian:
Remove corrupted tp_cdr rows before rerating

Upport #1737 from halliday

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/schema/tests/Kam/TrunksCdr/TrunksCdrRepositoryTest.php
+++ b/schema/tests/Kam/TrunksCdr/TrunksCdrRepositoryTest.php
@@ -22,6 +22,7 @@ class TrunksCdrRepositoryTest extends KernelTestCase
         $this->it_finds_one_by_callid();
         $this->it_finds_unparsedCalls();
         $this->it_resets_parsed_calls();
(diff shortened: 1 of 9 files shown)


schema/tests/Kam/TrunksCdr/TrunksCdrRepositoryTest.php
```diff
@@ -22,7 +22,8 @@ public function test_runner()
         $this->it_finds_one_by_callid();
         $this->it_finds_unparsedCalls();
         $this->it_resets_parsed_calls();
-        $this->it_resets_orphan_cgrids();
+        $this->it_get_cgrids_by_billableCalls();
+//        $this->it_resets_orphan_cgrids();
```
mmadariaga, line 26: Is this right?

Kaian: Nope, good catch, this was done while testing each of the functions individually because I was not sure what the source of the error was :smile:

PR #1831, opened 2022-06-24
Typescript prettier check

ironBatista:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


ironArt3mis:
Rebase completed.

Jenkinsfile
```diff
@@ -230,6 +230,7 @@ pipeline {
                     post {
                         success { notifySuccessGithub() }
                         failure { notifyFailureGithub() }
+                        always  { cleanWs() }
```
Kaian, line 233: Actually not needed. This test is run in a reuseNode agent, that means it shares the workspace with other tests (that also have reuseNode true in its agent section).

When the pipeline has ended, this shared workspace is cleaned once for all these reuseNode tests.

web/client/bin/test-lint
```diff
@@ -1,9 +1,22 @@
 #!/bin/bash
 
-pushd /opt/irontec/ivozprovider/web/client
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+[ -x $(which yarn) ] && YARN=$(which yarn) || YARN=$(which yarnpkg)
+
+echo 
+pushd "${SCRIPT_DIR}"/../
```
Kaian, line 7: We use hardcoded paths in all provider scripts, because they're designed to run in development enviroments, where paths and comands are fixed. No docker support is involved here, so this script is always located in the same path and yarn is always yarnkpkg.

Howerver, this is a valid enhancement, but I think It belongs to another PR (update test fixed paths). Also I think we may find a more clear approach for this project.

Kaian (changes requested):

Jenkinsfile
ironBatista: Hmm, now I got it, thanks, I wasn't sure if the process were like that, I'll remove ASAP.

web/client/bin/test-lint
ironBatista: Perfect, I'll roll back changes.

web/client/bin/test-lint
```diff
@@ -1,9 +1,17 @@
 #!/bin/bash
-
 pushd /opt/irontec/ivozprovider/web/client
     # Install required dependencies
     yarnpkg install
 
     # Run linter
     yarnpkg lint
+
+    CHANGED_FILES=$(git diff --name-only)
```
Kaian, line 9: I would add a dot at the end to only check current directory, because the check is run with others in parallel, this may catch differences in files in other paths.

    CHANGED_FILES=$(git diff --name-only .)

web/client/bin/test-lint
```diff
@@ -1,9 +1,17 @@
 #!/bin/bash
-
```
Kaian, line 2: Give my line back!

ironArt3mis:
Rebase completed.

Kaian (approved):

mmadariaga (approved):

ironArt3mis:
Rebase completed.

PR #1838, opened 2022-06-27
kamtrunks: remove duplicates from gws final avp

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Routing logic may add same carrierServer more than once (due to LCR).

This makes no sense as if a carrierServer fails for a call it will also fail for the same call some failover later.

#### Additional information
This PR modifies gateways AVP so that a carrier server is included only once (the first time).


ironArt3mis:
Rebase completed.

Kaian (approved):

Kaian (approved):

PR #1993, opened 2023-01-24
Make externallyRated a company setting

cruzccl:
#### Type Of Change 
- [ ] Small bug fix
- [x] Breaking change

#### Checklist:

#### Description
In Artemis _externallyRated_ setting was related to Carriers. In Halliday, _externallyRated_ is going to be related to Companies.

This way, calls from a given company may or may not be rated, no matter which carrier is used to place calls.

#### Additional information

This is a **breaking change** for platforms using a mix of _externallyRated_ carriers and non-_externallyRated_ ones for a given company. This was quite weird, as a call from a client A to a destination B may have rating details (price and cost) depending on the carrier used to place that specific call at that specific moment:
- If used carrier was marked as _externallyRated_: no rating details were calculated.
- Otherwise, rating details were calculated.

This was even more misleading as clients may call to to destinations not included in their pricing plans, as long as call was placed using a carrier marked as _externallyRated_. Depending on outgoing routing rules and carrier state in the moment of the call (up or down), a call may have price and may be validated against active pricing plan or not.

In Halliday _externallyRated_ setting is much simpler:
- If a client is marked with this setting, no destination will be validated against active pricing plan and call will have no rating details. These clients won't use any rating feature.
- If a client is not marked with this setting, it will only call to destinations incluided in active pricing plan and call will have rating details. These clients may use any rating feature.

### DIFF
--- a/schema/DoctrineMigrations/Version20230124171627.php
+++ b/schema/DoctrineMigrations/Version20230124171627.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Ivoz\Core\Infrastructure\Persistence\Doctrine\LoggableMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20230124171627 extends LoggableMigration
+{
+    public function getDescription(): string
+    {
+        return 'New value none for Companies.billingMethod';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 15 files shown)


schema/DoctrineMigrations/Version20230124171627.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20230124171627 extends AbstractMigration
```
mmadariaga, line 13: `use Ivoz\Core\Infrastructure\Persistence\Doctrine\LoggableMigration;`

```suggestion
final class Version20230124171627 extends LoggableMigration
```

mmadariaga (changes requested):

mmadariaga (approved):

PR #2025, opened 2023-02-03
Review MatchLists rest API ACLS

Kaian:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
MatchLists are used in brand (aka generic matchlists) and client portals.
Brand MatchLists have brandId with value and companyId set to NULL.
Client MatchLists have brandId set to NULL and companyId with value.

Brand Administrators can only work (read/write/delete) with their own Brand MatchLists.
Client Administrators can read Brand MatchLists, but can only work (read/write/delete) with their own Client MatchLists.

Same logic applies to its child entity MatchListPattern, although not implemented here.

### DIFF
--- a/library/Ivoz/Provider/Domain/Model/MatchList/MatchListDto.php
+++ b/library/Ivoz/Provider/Domain/Model/MatchList/MatchListDto.php
@@ -2,8 +2,21 @@
 
 namespace Ivoz\Provider\Domain\Model\MatchList;
 
+use Ivoz\Api\Core\Annotation\AttributeDefinition;
+
+/** @psalm-suppress UnusedProperty */
 class MatchListDto extends MatchListDtoAbstract
 {
+    /**
+     * @var string
+     * @AttributeDefinition(
+     *     type="bool",
+     *     writable=false,
+     *     description="Generic Match List"
+     * )
+     */
+    private $generic;
+
     /**
      * @inheritdoc
      * @codeCoverageIgnore
@@ -11,16 +24,17 @@ class MatchListDto extends MatchListDtoAbstract
     public static function getPropertyMap(string $context = '', string $role = null): array
     {
         if ($context === self::CONTEXT_COLLECTION) {
-            return [
+            $response =  [
                 'id' => 'id',
-                'name' => 'name'
+                'name' => 'name',
             ];
+        } else {
+            $response = parent::getPropertyMap(...func_get_args());
         }
-
-        $response = parent::getPropertyMap(...func_get_args());
 
         if ($role === 'ROLE_COMPANY_ADMIN') {
             unset($response['companyId']);
+            $response['generic'] = 'generic';
         }
 
         if ($role === 'ROLE_BRAND_ADMIN') {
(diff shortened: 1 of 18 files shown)


library/Ivoz/Provider/Domain/Model/MatchList/MatchListDto.php
```diff
@@ -31,6 +45,28 @@ public static function getPropertyMap(string $context = '', string $role = null)
...
+            $role
+        );
+
+        if ($context === self::CONTEXT_COLLECTION) {
+            if ($role === 'ROLE_COMPANY_ADMIN') {
+                $editable = $this->getCompanyId() != null;
+                $this->setEditable($editable);
+                $response['editable'] = $editable;
+            }
+        }
```
Kaian, line 64: Psalm complains for attribute `$editable` not being used, and other classes have getters and setters for custom fields. I liked the idea but I was not sure what would be the best way to implement it.

web/rest/brand/features/provider/matchList/putMatchList.feature
```diff
@@ -24,3 +24,25 @@ Feature: Update match lists
           "id": 3
       }
     """
+
+  Scenario: Match list updates ignores company values
+    Given I add Brand Authorization header
+    When I add "Content-Type" header equal to "application/json"
```
mmadariaga, line 30: Align this please

web/rest/brand/features/provider/matchList/putMatchList.feature
```diff
@@ -24,3 +24,25 @@ Feature: Update match lists
...
+      }
+    """
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json; charset=utf-8"
+    And the JSON should be like:
+    """
+     {
+          "name": "updatedMatchlist",
+          "id": 3
```
mmadariaga, line 46: Shouldn't you check company here?

mmadariaga (changes requested):

web/rest/brand/features/provider/matchList/putMatchList.feature
Kaian: I think I can't actually test this through rest API because that field is not part of the spec :-\

Kaian: I think I can't actually test this through rest API because that field is not part of the spec :-\

web/rest/brand/features/provider/matchList/putMatchList.feature
Kaian: I have formatted all matchList api tests using [this script](https://pypi.org/project/reformat-gherkin/). Is the only one I've found that is able to align format to the right. 

I'll add it to the pipeline in a later pull request, with all files updated.

web/rest/brand/features/provider/matchList/putMatchList.feature
Kaian: Test removed.

The value of company is not provided in the response, so it can not be checked. The matchlist could be retrieved again and assume the value has not changed because GET access_control still works (it filters out entries with companyId filled),  but don't like it too much either.

mmadariaga (approved):

PR #2027, opened 2023-02-03
Update InvoiceTemplate rest API tests with extra scenarios

Kaian:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Added scenarios to check Global InvoiceTemplates can not be removed or edited from brand API.

### DIFF
--- a/web/rest/brand/features/provider/invoiceTemplate/putInvoiceTemplate.feature
+++ b/web/rest/brand/features/provider/invoiceTemplate/putInvoiceTemplate.feature
@@ -15,8 +15,7 @@ Feature: Update invoice templates
           "description": "updated something",
           "template": "body v2",
           "templateHeader": "header v2",
-          "templateFooter": "footer v2",
-          "id": 1
+          "templateFooter": "footer v2"
       }
     """
     Then the response status code should be 200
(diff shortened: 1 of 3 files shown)


web/rest/brand/features/provider/invoiceTemplate/putInvoiceTemplate.feature
```diff
@@ -34,3 +33,19 @@ Feature: Update invoice templates
...
+    """
+      {
+          "name": "Updated",
+          "description": "updated global invoice template",
+          "template": "global body updated",
+          "templateHeader": "global header updated",
+          "templateFooter": "global footer updated"
+      }
+    """
+    Then the response status code should be 400
```
Kaian, line 51: Note this is a 400 instead of 403 response, because entity sanitize is throwing an exception that avoid this scenario, so API acl is not actually being used to avoid this update.

mmadariaga: API require entity instance before they can be run. You can add response code to the sanitize exception if you want.

`throw new \DomainException('bla bla bla ', 403);`

mmadariaga (approved):

PR #2059, opened 2023-02-10
Create client error

mmadariaga:
Allow empty company invoicing data

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:


cruzccl (changes requested):
- Klear site shows these fields as required (avoid saving existing elements with empty strings in these fields).
- Cannot see any of these fields in new web brand portals (for a brand with invoices feature enabled), I get TODO FIELD TYPE.

fmendozzac (approved):

Kaian (approved):

web/rest/brand/config/api/raw/provider.yml
```diff
@@ -342,6 +342,7 @@ Ivoz\Provider\Domain\Model\WebPortal\WebPortal:
       - urlType
 Ivoz\Provider\Domain\Model\WebPortal\Logo: ~
 
+Ivoz\Provider\Domain\Model\Company\Invoicing: ~
```
rbatistadev, line 345: This change will permit every operation over Invoicing Entity, is this the intention? or we want to load only Invoicing Model to use on some parent Entity?

```suggestion
Ivoz\Provider\Domain\Model\Company\Invoicing:
  itemOperations: []
  collectionOperations: []
```

rbatistadev (approved):

web/rest/brand/config/api/raw/provider.yml
mmadariaga: Invoicing is not an entity, just a VO. Like Logo, multilang descriptions, etc.

PR #2211, opened 2023-03-28
Fixed residential device creation

mmadariaga:
- Filter company selector so that it only shows residential clients
- Added missing fields

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:


Kaian (changes requested):
If would be nice if Company field is required. Right now a backend exception is rising when this field is leaving empty
```
 getCompany value is null, but non null value was expected.
```

Kaian (approved):
Sweet, also changing the position of the Company field is a nice addition :+1:

PR #2214, opened 2023-03-28
Portal brand destination rate when timing is always it fails

rbatistadev:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Portal brand destination rate when fix fails when timing is always


library/Ivoz/Provider/Domain/Model/RatingPlan/RatingPlanDto.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Ivoz\Provider\Domain\Model\RatingPlan;
 
+use function PHPUnit\Framework\isNull;
```
mmadariaga, line 5: nope

mmadariaga (changes requested):

PR #2253, opened 2023-05-30
Carrier calculate cost

mmadariaga:
Carriers without billing feature must have calculate cost equal to false

#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/library/Ivoz/Provider/Domain/Model/Carrier/Carrier.php
+++ b/library/Ivoz/Provider/Domain/Model/Carrier/Carrier.php
@@ -36,6 +36,14 @@ class Carrier extends CarrierAbstract implements CarrierInterface
      */
     protected function sanitizeValues(): void
     {
+        $billingEnabled = $this
+            ->getBrand()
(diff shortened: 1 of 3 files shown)


library/Ivoz/Provider/Domain/Model/Carrier/Carrier.php
```diff
@@ -36,6 +36,14 @@ public function getId(): ?int
      */
     protected function sanitizeValues(): void
     {
+        $billingEnabled = $this
+            ->getBrand()
+            ->hasFeatureByIden('billing');
```
Kaian, line 41: We should add constants for these someday. They exist for ids which seems much worse than hardcoding idens.

Kaian (approved):

library/Ivoz/Provider/Domain/Model/Carrier/Carrier.php
mmadariaga: I agree. Added to backlog

PR #2262, opened 2023-06-08
Billable Calls exporters

mmadariaga:
Added export to CSV button in billable calls

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


web/portal/platform/src/entities/BillableCall/Action/Export.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { useStoreActions } from 'store';
+
+import BillableCall from '../BillableCall';
+
+const Export: ActionFunctionComponent = (props: MultiSelectActionItemProps) => {
+  const { variant = 'icon' } = props;
+
+  const [open, setOpen] = useState(false);
+  const [error, setError] = useState(false);
+  const [errorMsg, setErrorMsg] = useState<string>('');
```
rbatistadev, line 31: Is this hint required?, I mean, when you set the initial value in useState, automatically infers such value from it.

rbatistadev (changes requested):
Pr review

web/portal/platform/src/entities/BillableCall/Action/Export.tsx
mmadariaga: You're right. I'll change it

PR #2314, opened 2023-07-01
Portal platform not found snackbar when inital load at list

rbatistadev:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Fixed path literal in DDi entity from platform portal, and exposed its collection in the API.

### DIFF
--- a/web/portal/platform/src/entities/BillableCall/BillableCall.tsx
+++ b/web/portal/platform/src/entities/BillableCall/BillableCall.tsx
@@ -121,6 +121,11 @@ const BillableCall: EntityInterface = {
     delete: false,
   },
   customActions: Actions,
+  foreignKeyResolver: async () => {
+    const module = await import('./ForeignKeyResolver');
+
+    return module.default;
+  },
   View: async () => {
     const module = await import('./View');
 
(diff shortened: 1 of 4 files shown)


web/rest/platform/config/api/raw/provider.yml
```diff
@@ -43,7 +43,8 @@ Ivoz\Provider\Domain\Model\BillableCall\BillableCall:
 
 Ivoz\Provider\Domain\Model\Ddi\Ddi:
   itemOperations: []
-  collectionOperations: []
+  collectionOperations:
```
mmadariaga, line 46: why do we need to expose ddis at platform level?

mmadariaga (changes requested):

web/rest/platform/config/api/raw/provider.yml
rbatistadev: Because in the external calls list, there is a fkresolver pointing to Ddi's of API platform, and when i tried to get Ddi's in the initial request where the list is being loaded failed because of two reasons, the Ddi entity path was with a typo and the endpoint wasn't published.

mmadariaga: But it's not shown anywhere so, just skip that call

mmadariaga (approved):

PR #2493, opened 2024-01-29
Many proxy users

mmadariaga:
Allow to add multiple proxy users in the UI

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/web/rest/platform/config/api/raw/provider.yml
+++ b/web/rest/platform/config/api/raw/provider.yml
@@ -355,15 +355,19 @@ Ivoz\Provider\Domain\Model\ProxyTrunksRelBrand\ProxyTrunksRelBrand:
     access_control: '"ROLE_SUPER_ADMIN" in roles && user.hasAccessPrivileges(_api_resource_class, request.getMethod())'
 
 Ivoz\Provider\Domain\Model\ProxyUser\ProxyUser:
-  itemOperations:
-    get: ~
-    put: ~
-  collectionOperations:
-    get: ~
+  itemOperations: ~
+  collectionOperations: ~
   attributes:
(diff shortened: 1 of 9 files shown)


web/rest/platform/config/api/raw/provider.yml
```diff
@@ -355,15 +355,19 @@ Ivoz\Provider\Domain\Model\ProxyTrunksRelBrand\ProxyTrunksRelBrand:
...
-    get: ~
+  itemOperations: ~
+  collectionOperations: ~
   attributes:
     access_control: '"ROLE_SUPER_ADMIN" in roles && user.hasAccessPrivileges(_api_resource_class, request.getMethod())'
     order:
       name: ASC
+    write_access_control:
+      ROLE_SUPER_ADMIN:
+        raw: 'this.hasChanged("name") == false || this.getId() != 1'
```
Kaian, line 366: Not sure why this condition exists, it should have the same access controls as ProxyTrunks entity above :thinking: Maybe this condition should also be present in the ProxyTrunks endpoint?

web/rest/platform/features/provider/proxyUser/removeProxyUser.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Manage proxy users
+  In order to manage proxy trunks rel brands
```
Kaian, line 2: In order to manage proxy users

Kaian (changes requested):

cruzccl (approved):

web/rest/platform/config/api/raw/provider.yml
mmadariaga: Those are the rules: default proxy user's name cannot be changed. It wasn't possible before either (but only at UI level).

I may agree with the second part but that would be part of another issue

Kaian (approved):

PR #2502, opened 2024-01-31
Fixed registration channel filters

rbatistadev:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Fixed RegistrationChannelResolver when clients send, ambiguous filtering

### DIFF
--- a/library/Ivoz/Provider/Application/Service/ActiveCall/RegistrationChannelResolver.php
+++ b/library/Ivoz/Provider/Application/Service/ActiveCall/RegistrationChannelResolver.php
@@ -2,7 +2,6 @@
 
 namespace Ivoz\Provider\Application\Service\ActiveCall;
 
-use Ivoz\Provider\Domain\Model\Administrator\Administrator;
 use Ivoz\Provider\Domain\Model\Administrator\AdministratorRepository;
 use Ivoz\Provider\Domain\Service\ActiveCall\AssertAccessGranted;
 
@@ -25,37 +24,29 @@ class RegistrationChannelResolver
      */
     public function execute(?array $filters, array $tokenPayload): string
     {
-        $defaultCriteria = $this->getDefaultCriteria($tokenPayload);
-
-        $role = $tokenPayload['roles'][0];
-
-        $trunksRoles = [
-            Administrator::ROLE_BRAND_ADMIN,
-            Administrator::ROLE_SUPER_ADMIN
-        ];
-
-        if (in_array($role, $trunksRoles)) {
-            $direction = $filters['direction'] ?? null;
-            $cr = $filters['cr'] ?? null;
-            $dp = $filters['dp'] ?? null;
-
-            if ($direction === INBOUND && is_null($cr)) {
-                $cr = '*';
-            }
-
-            if ($direction === OUTBOUND && is_null($cr)) {
-                $dp = '*';
-            }
-
-            $defaultCriteria['trunks']['b'] = $filters['b'] ?? null;
-            $defaultCriteria['trunks']['c'] = $filters['c'] ?? null;
-            $defaultCriteria['trunks']['cr'] = $cr;
-            $defaultCriteria['trunks']['dp'] = $dp;
+        if (!$this->filtersMakeSense($filters)) {
+            return "trunks:-1";
         }
 
-        return $this->criteriaToString(
+        $criteria = $defaultCriteria = $this->getDefaultCriteria(
+            $tokenPayload
+        );
+        $filteringTrunks = isset($defaultCriteria['trunks']);
+
+        if ($filteringTrunks) {
+            $criteria = $this->getTrunksCriteria(
+                $filters,
+                $defaultCriteria
+            );
+        }
+
+        $this->assertAccessGranted->execute(
             $tokenPayload,
-            $defaultCriteria
+            $criteria
+        );
+
+        return $this->toString(
+            $criteria
         );
     }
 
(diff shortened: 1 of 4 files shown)


web/rest/platform/features/my/activeCalls/getActiveCallsFilters.feature
```diff
@@ -11,6 +11,20 @@ Feature: Retrieve active calls filters
...
+      And I send a "GET" request to "/my/active_calls/realtime_filter?b=1&c=2&dp=4"
+     Then the response status code should be 200
+      And the response should be in JSON
+      And the header "Content-Type" should be equal to "application/json"
+      And the JSON should be equal to:
+      """
+      {
+          "criteria": "trunks:b1:c2:dp4:*"
       }
       """
```
Kaian, line 30: Can you also add tests for filters for a given Carrier?
Can you also add tests for DdiProvider&Direction and Carrier&Direction ? P.e. DdiProvider and direction = inbound and DdiProvider and direction = outbound.

This tests will also be interesting for the other APIs

Kaian (changes requested):

web/rest/brand/features/my/activeCalls/getActiveCallsFilters.feature
```diff
@@ -11,6 +11,76 @@ Feature: Retrieve active calls filters
...
+    And I send a "GET" request to "/my/active_calls/realtime_filter?cr=2"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON should be equal to:
+      """
+      {
+          "criteria": "trunks:b1:*:cr2:*"
+      }
+      """
```
Kaian, line 58: Duplicate test?

web/rest/platform/features/my/activeCalls/getActiveCallsFilters.feature
```diff
@@ -11,6 +11,76 @@ Feature: Retrieve active calls filters
...
+    And I send a "GET" request to "/my/active_calls/realtime_filter?cr=2"
+    Then the response status code should be 200
+    And the response should be in JSON
+    And the header "Content-Type" should be equal to "application/json"
+    And the JSON should be equal to:
+      """
+      {
+          "criteria": "trunks:*:*:cr2:*"
+      }
+      """
```
Kaian, line 58: Also duplicated

Kaian (changes requested):

Kaian:
We also need additional tests for direction only filtering.
There trunks realtime keys have the format:

```
trunks:<brand tag>:<company tag>:<carrier tag|ddi provider tag>:<call-id>
```

Keys with carrier are **outbound** calls and calls with ddi provider are **indbound** calls.

If you want to get only outbound calls, we need a filter that matches all carriers, pe:

```
trunks:*:*:cr*:*
```

If you want to get only inbound calls, we need a filter that matches all ddi providers, pe:
```
trunks:*:*:dp*:*
```

Take into account that carrier and ddi provider filters should have more priority that direction. If you want outbound calls for carrier #1, the key should be:

```
trunks:*:*:cr1:*
```

In that case, the direction does not provide any extra filtering, because all calls from Carrier #1 are outbound. 

If the user supplies Carrier and Direction inbound (which makes no sense), you can give an error, just like filtering by Carrirer and DDI provider at the same time.

library/Ivoz/Provider/Application/Service/ActiveCall/RegistrationChannelResolver.php
```diff
@@ -39,12 +39,15 @@ public function execute(?array $filters, array $tokenPayload): string
...
-            if ($direction === OUTBOUND && is_null($cr)) {
-                $dp = '*';
+            $canFilterByDirection = !is_null($direction) && is_null($cr) && is_null($dp);
+            if ($canFilterByDirection) {
+                if ($direction === 'inbound') {
+                    $cr = 'cr*';
+                }
+
+                if ($direction === 'outbound') {
+                    $dp = 'dp*';
```
Kaian, line 49: Actually the other way around. DDI Providers are used for inbound calls, Carrier for outbound calls. If you filter for outbound call you want to match all carriers (`cr*`);

Kaian (changes requested):

web/rest/platform/features/my/activeCalls/getActiveCallsFilters.feature
```diff
@@ -11,6 +11,90 @@ Feature: Retrieve active calls filters
...
+      And I send a "GET" request to "/my/active_calls/realtime_filter?dp=1&direction=outbound"
+     Then the response status code should be 200
+      And the response should be in JSON
+      And the header "Content-Type" should be equal to "application/json"
+      And the JSON should be equal to:
+      """
+      {
+          "criteria": "trunks:*:*:dp1:*"
+      }
+      """
```
Kaian, line 72: I would return error in these cases: 

```
"criteria": "trunks:-1"
```

While filtering for DDI Provider 1 all calls will be inbound. If you add a direction outbound to the filter, no call should be displayed, but with this response the direction filter is being ignored and Inbound DDI Provider calls will be displayed. Same applies to Carrier and outbound calls: if both filters are set, no call should be displayed.

There was a reference to this in my note:

> If the filter has Carrier and Direction inbound or DDI Provider and Direction outbound (which makes no sense), you can give an error, just like filtering by Carrirer and DDI provider at the same time.

Kaian (changes requested):

mmadariaga (approved):

Kaian (changes requested):
Filter when no parameters are passed is not correct . Add this test to **platform** to verify it:

```
  Scenario: Retrieve platform active calls filter with no parameters
    Given I add Authorization header
     When I add "Accept" header equal to "application/json"
      And I send a "GET" request to "/my/active_calls/realtime_filter"
     Then the response status code should be 200
      And the response should be in JSON
      And the header "Content-Type" should be equal to "application/json"
      And the JSON should be equal to:
      """
      {
          "criteria": "trunks:*:*:*:*"
      }
      """

This also happens in brand API. Add this test to **brand** to verify it:

```
  Scenario: Retrieve brand active calls filter json with no parameters
    Given I add Brand Authorization header
     When I add "Accept" header equal to "application/json"
      And I send a "GET" request to "/my/active_calls/realtime_filter"
     Then the response status code should be 200
      And the response should be in JSON
      And the header "Content-Type" should be equal to "application/json"
      And the JSON should be equal to:
      """
      {
          "criteria": "trunks:b1:*:*:*"
      }
      """
```

Kaian (approved):

PR #2505, opened 2024-02-01
Enhanced tz conversion on API impersonation

mmadariaga:
From now on internal admins will inherit timezone from their parent admin (the first one that is not internal in the impersonation chain). On the other hand, "real admin" impersonation will keep it's timezone.

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


Kaian (changes requested):
Looks like this breaks one of the Options is Platform Main Operators list:

image

Kaian (approved):

PR #2508, opened 2024-02-06
Access methods to domains in client api should be similar

rbatistadev:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Updated domain access methods for:
- Terminal
- Residential devices
- Retail Account


library/Ivoz/Provider/Domain/Model/ResidentialDevice/ResidentialDeviceDto.php
```diff
@@ -106,10 +106,13 @@ public static function getPropertyMap(string $context = '', string $role = null)
                 $response['rtpEncryption'] = 'rtpEncryption';
                 $response['multiContact'] = 'multiContact';
             }
-        } else {
-            $response = parent::getPropertyMap(...func_get_args());
+
+            return $response;
```
mmadariaga, line 110: You're skipping filterFieldsForCompanyAdmin call with this early return

web/rest/client/features/provider/retailAccount/getRetailAccountStatus.feature
```diff
@@ -48,7 +48,12 @@ Feature: Retrieve retail accounts status
           "description": "",
           "directConnectivity": "no",
           "id": 1,
-          "domainName": "retail.irontec.com",
+          "domain": {
```
mmadariaga, line 51: This is not a collection response and it's out of the scope of this PR

mmadariaga (changes requested):

web/rest/brand/features/provider/domain/getDomains.feature
```diff
@@ -0,0 +1,36 @@
...
+              "id": 5
+          }
+      ]
+      """
+
+  Scenario: Retrieve certain domain json
+    Given I add Authorization header
+     When I add "Accept" header equal to "application/json"
+      And I send a "GET" request to "domains/3"
+     Then the response status code should be 404
```
Kaian, line 36: Why is this 404 if the collection contains a Domain with id 3?

library/Ivoz/Provider/Infrastructure/Persistence/Doctrine/CompanyDoctrineRepository.php
```diff
@@ -59,6 +59,32 @@ function ($row): int {
...
+            ->getQuery()
+            ->getScalarResult();
+
+        return array_map(
+            function ($row): int {
+                return (int) $row['domain'];
+            },
+            $result
+        );
+    }
```
Kaian, line 86: This makes me think that Domain should have a brandId and companyId fields, because this query should be part of Domain Repository. This is currently not possible, so this is the only approach to get Brand Domains.

web/rest/brand/features/provider/domain/getDomains.feature
mmadariaga: Because there is no `/domains/{id}` endpoint, just `/domains`

mmadariaga (approved):

PR #2512, opened 2024-02-09
Review UsersCdrs API exposed fields

Kaian:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Ensure brand and company are not exposed in APIs that should not be aware of those fields.

### DIFF
--- a/library/Ivoz/Provider/Domain/Model/UsersCdr/UsersCdrDto.php
+++ b/library/Ivoz/Provider/Domain/Model/UsersCdr/UsersCdrDto.php
@@ -21,6 +21,16 @@ class UsersCdrDto extends UsersCdrDtoAbstract
             $response = parent::getPropertyMap(...func_get_args());
         }
 
+        if ($role === 'ROLE_BRAND_ADMIN') {
+            unset($response['brandId']);
+        } elseif ($role === 'ROLE_COMPANY_ADMIN') {
+            unset($response['brandId']);
+            unset($response['companyId']);
+        } elseif ($role === 'ROLE_COMPANY_USER') {
+            unset($response['brandId']);
+            unset($response['companyId']);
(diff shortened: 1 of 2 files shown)


library/Ivoz/Provider/Domain/Model/UsersCdr/UsersCdrDto.php
```diff
@@ -17,10 +17,28 @@ public static function getPropertyMap(string $context = '', string $role = null)
...
+                    'friendId' => 'friend',
+                    'extensionId' => 'extension',
+                ];
+            }
         } else {
             $response = parent::getPropertyMap(...func_get_args());
         }
 
+        if ($role === 'ROLE_BRAND_ADMIN') {
+            unset($response['brandId']);
```
Kaian, line 33: This entity is not currently exposed in Brand API (due to an error), so this condition is debatable

library/Ivoz/Provider/Domain/Model/UsersCdr/UsersCdrDto.php
```diff
@@ -17,10 +17,28 @@ public static function getPropertyMap(string $context = '', string $role = null)
...
+        } elseif ($role === 'ROLE_COMPANY_ADMIN') {
+            unset($response['brandId']);
+            unset($response['companyId']);
+        } elseif ($role === 'ROLE_COMPANY_USER') {
+            unset($response['brandId']);
+            unset($response['companyId']);
+        }
```
Kaian, line 40: This only applies to non-collection contexts of user API. There is no detailed view of UsersCdrs in that API right now, but it will be required sooner than later, so this is also debatable

cruzccl (approved):

PR #2711, opened 2024-07-26
Add fields to recordings

danigargar:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/web/rest/client/features/provider/recording/getRecording.feature
+++ b/web/rest/client/features/provider/recording/getRecording.feature
@@ -14,16 +14,36 @@ Feature: Retrieve recordings
       And the JSON should be equal to:
       """
       [
-          {
-              "callid": "7602fd7f-4153-4475-9100-d89ff70cdf76",
-              "calldate": "2017-01-05 01:15:15",
-              "type": "ondemand",
-              "duration": 3,
-              "caller": "34946002020",
-              "callee": "34946002021",
-              "recorder": null,
-              "id": 1
-          }
+            {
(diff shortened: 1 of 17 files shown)


rbatistadev:
Update API tests, and Cypress tests please

rbatistadev (changes requested):
Cypress tests and API tests please

danigargar:
> Cypress tests and API tests please

This PR is part of an epic.
Cypress tests do not fit in this PR, as there are no modifications in the frontend. What should I test?
As for the backend, something similar happens: no real behavior is altered in either the backend or the frontend. Fields are added, but without any use.

I will add Cypress tests in the following tasks, which require adding a section for user/Recordings. Backend tests will also fit better here, as the behavior of some endpoints will be modified and new ones will be created.

web/rest/client/features/provider/recording/getRecording.feature
```diff
@@ -49,6 +49,8 @@ Feature: Retrieve recordings
               "fileSize": 4280,
               "mimeType": "audio/mpeg; charset=binary",
               "baseName": "7602fd7f-4153-4475-9100-d89ff70cdf76.0.mp3"
-          }
+          },
+          "ddi": null,
+          "user": null
```
rbatistadev, line 54: We should test that, these values can be updated, and this can be performed in the RecordingStub, having elements to null does not test if it is a relation to the required entities

rbatistadev (approved):

PR #2713, opened 2024-08-01
Store ddi or user in recordings

danigargar:
Added logic in recordings microservice for setting user or ddi for new records.

#### Type Of Change 
- [ ] Small bug fix
- [x ] New feature or enhancement
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/microservices/recordings/src/Service/Encoder.php
+++ b/microservices/recordings/src/Service/Encoder.php
@@ -9,6 +9,7 @@ use Ivoz\Kam\Domain\Model\UsersCdr\UsersCdrInterface;
 use Ivoz\Kam\Domain\Model\UsersCdr\UsersCdrRepository;
 use Ivoz\Provider\Domain\Model\Ddi\DdiRepository;
 use Ivoz\Provider\Domain\Model\Recording\RecordingDto;
+use Ivoz\Provider\Domain\Model\Recording\RecordingInterface;
 use Symfony\Bridge\Monolog\Logger;
 use Symfony\Component\Process\Process;
 
@@ -134,7 +135,7 @@ class Encoder
 
             foreach ($kamAccCdrs as $kamAccCdr) {
                 if ($kamAccCdr instanceof TrunksCdrInterface) {
-                    $type = 'ddi';
+                    $type = RecordingInterface::TYPE_DDI;
                     $direction = $kamAccCdr->getDirection();
 
                     if ($direction == 'outbound') {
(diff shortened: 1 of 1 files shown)


ironArt3mis (changes requested):
Functional review required

microservices/recordings/src/Service/Encoder.php
```diff
@@ -267,6 +271,20 @@ public function processAction(): void
                     ->setRecordedFileBaseName($baseName)
                     ->setRecordedFilePath($convertMp3);
 
+
+                $isDdi = $type === self::DDI && is_object($ddi);
+                $isUser = $type === self::ONDEMAND && is_object($user);
```
rbatistadev, line 276: $ddi and $user are actually objects or possible null values, what about?:

```suggestion
                $isDdi = $type === self::DDI && !is_null($ddi);
                $isUser = $type === self::ONDEMAND && !is_null($user);
```

microservices/recordings/src/Service/Encoder.php
```diff
@@ -23,6 +23,8 @@ class Encoder
      * Recording created this seconds ago will be ignored
      */
     const RECORDING_AGE_MIN = 10;
+    protected const ONDEMAND = 'ondemand';
+    protected const DDI = 'ddi';
```
Kaian, line 27: 2 cents, there are contacts with the same values in the Recording entity interface:

https://github.com/irontec/ivozprovider/blob/5ad4bb273592b11e6f23d2a80f727f433402c4c1/library/Ivoz/Provider/Domain/Model/Recording/RecordingInterface.php#L17-L21

microservices/recordings/src/Service/Encoder.php
danigargar: Those are "possible" undefined. I mean, if $type === self::DDI, then I'm sure $ddi is an object.
The same way., I'm sure that when $isUser === true $user is an object.

This check is only meant to provide some documentation for a pending (and scheduled) full refactor of this code. I rather keep the "is_object" because is more assertive but I could change it to is_null or removing it altogether, if you insist. Though I'm not sure which strategy is best in absence of tests and in case of a later modification: 
1) removing this, which could end with a fatal error.
2) keeping a check, which could result in a poorly functioning, but functional script (as in "it works, but stops storing the user or the ddi").

microservices/recordings/src/Service/Encoder.php
danigargar: Great! Didn't notice that.
Thanks!

rbatistadev (approved):

cruzccl (approved):

ironArt3mis (approved):

PR #2720, opened 2024-08-06
Unify web portals testing scripts

Kaian:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:
- [X] Commits are split per component (schema, portal/platform, kamusers, agis, ..)

#### Description

Before this change, frontend test scripts were located in platform portal bin dir, while other portals have symlinks to that location.
This PR introduce following changes:
1. Moves those scripts so there is a common script directory for all portals and updates existing symlinks
2. Update scripts to properly detect where the portal base directory is located
3. Removes the duality between yarn and yarnpkg commands (no longer exists in Debian 12)
4. Unifies the codestyle of all tests


web/portal/scripts/test-sync-api-spec
```diff
@@ -0,0 +1,21 @@
...
+if [ -z "$API_TYPE" ]; then
+    echo "Usage: $0 [platform|brand|client|user]"
+    exit 1
+fi
+
+# Move to webapp dir
+cd $(dirname "$(realpath -se "$0")")/..
+
+# Generate requested API spec
+/opt/irontec/ivozprovider/web/rest/$API_TYPE/bin/console api:swagger:export > ./cypress/fixtures/apiSpec.json
```
danigargar, line 13: This is only a suggestion:
`./bin/console api:swagger:export > ./cypress/fixtures/apiSpec.json
`
Should work fine and we could get rid of the API_TYPE check.

danigargar (changes requested):

web/portal/scripts/test-sync-api-spec
Kaian: The current path is the directory of the portal, and the bin/console must be run in the directory of the rest api application, that's why the full path is specified

p.e: to generate /opt/irontec/ivozprovider/web/portal/platform/cypress/fixtures/apiSpec.json the bin/console script has to be /opt/irontec/ivozprovider/web/rest/platform/bin/console

danigargar: Of course you are right

danigargar (approved):

PR #2725, opened 2024-08-12
Recordings in user portal

danigargar:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/web/portal/user/cypress/e2e/Recording/Recording.cy.ts
+++ b/web/portal/user/cypress/e2e/Recording/Recording.cy.ts
@@ -1,0 +1,60 @@
+import RecordingCollection from '../../fixtures/Recording/getCollection.json';
+import RecordingItem from '../../fixtures/Recording/getItem.json';
+
+describe('CallForwardSetting', () => {
+  beforeEach(() => {
+    cy.prepareGenericPactInterceptors('User-Recording');
+    cy.before();
+
+    cy.get('svg[data-testid="SettingsVoiceIcon"]').first().click();
+
+    cy.get('table').should('contain', RecordingCollection.body[0].duration);
+  });
+
+  it('View CallForwardSetting', () => {
+    const recordingId = RecordingItem.body.id;
+    cy.usePactIntercept(
+      {
+        method: 'GET',
+        url: `**api/user/recordings/${recordingId}`,
+        response: RecordingItem.body,
+      },
+      `getRecording-${recordingId}`
+    );
+
+    cy.usePactIntercept(
+      {
+        method: 'GET',
+        url: '**api/user/recordings/2/recordedfile',
+        response: '',
+      },
+      'getRecordedFile-2'
+    );
+
+    cy.get('svg[data-testid="PanoramaIcon"]').first().click();
+
+    cy.usePactWait(`getRecording-${recordingId}`)
+      .its('response.statusCode')
+      .should('eq', 200);
+
+    cy.get('input[name=callee]');
+  });
+
+  it('delete CallForwardSetting', () => {
+    cy.intercept('DELETE', '**api/user/recordings/*', {
+      statusCode: 204,
+    }).as('deleteRecording');
+
+    cy.get('table svg[data-testid="DeleteIcon"]').first().click();
+
+    cy.contains('Remove element');
+    cy.get('div.MuiDialog-container button')
+      .filter(':visible')
+      .contains('Yes, delete it')
+      .click({ force: true });
+
+    cy.usePactWait('deleteRecording')
+      .its('response.statusCode')
+      .should('eq', 204);
+  });
+});
--- a/web/rest/user/config/api/raw/provider.yml
+++ b/web/rest/user/config/api/raw/provider.yml
@@ -386,3 +386,44 @@ Ivoz\Provider\Domain\Model\Fax\Fax:
     swagger_context:
       required:
         - company
+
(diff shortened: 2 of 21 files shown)


ironArt3mis (changes requested):
Functional review required

Kaian:
Note: Tests in this pull request are actually passing, and the pr-head failing test will disappear as soon as the feature branch is rebased

web/portal/user/cypress/e2e/Recording/Recording.cy.ts
```diff
@@ -0,0 +1,59 @@
...
+        url: '**api/user/recordings/*',
+        response: RecordingItem.body,
+      },
+      'getRecording-1'
+    );
+
+    cy.usePactIntercept(
+      {
+        method: 'GET',
+        url: '**api/user/recordings/*/recordedfile',
```
rbatistadev, line 27: this could work for general programming in recording files, but here, I'm pretty sure that you know the ID of the recording, so it's a better idea to use the ID of this use case than a mask for every request

rbatistadev: PD: the id is "id": 2, and I recommend writing the string for the interceptor, as follows:
'getRecordedFile-{ID}'

rbatistadev:
```suggestion
        export const putCompany = () => {
  cy.usePactIntercept(
    {
      method: 'GET',
      url: '**/api/brand/companies/1',
      response: { ...CompaniesItem },
    },
    'getCompanies-1'
  );

  cy.usePactIntercept(
    {
      method: 'PUT',
      url: `**/api/brand/companies/${editCompanies.response.body.id}`,
      response: editCompanies.response,
    },
    'editCompanies'
  );

  cy.get('svg[data-testid="EditIcon"]').first().click();

  testPbx({ ...editCompanies, mode: MODE.EDIT });

  cy.get('header').should('contain', 'Virtual PBXs');

  cy.usePactWait(['editCompanies'])
    .its('response.statusCode')
    .should('eq', 200);
};
```

web/rest/user/config/api/raw/provider.yml
```diff
@@ -386,3 +386,44 @@ Ivoz\Provider\Domain\Model\Fax\Fax:
...
+      path: '/recordings/{id}'
+      requirements:
+        id: '\d+'
+    delete: ~
+  collectionOperations:
+    get: ~
+    get_multiple_recorderded_files:
+      method: 'GET'
+      path: '/recordings/recorded_files_zip'
+      route_name: 'get_multiple_recorderded_files'
```
rbatistadev, line 403: Normally we try to resemble the names of route_name and path

    get_recorderded_files_zip:
      method: 'GET'
      path: '/recordings/recorded_files_zip'
      route_name: 'get_recorderded_files_zip'

web/rest/user/src/Controller/Provider/RecordedFilesZipAction.php
```diff
@@ -0,0 +1,44 @@
...
+        $ids = explode(
+            ',',
+            (string) $request->query->get('_recordingIds')
+        );
+
+        $zipFilePath = $this
+            ->zipRecordingFiles
+            ->execute($ids);
+
+        // This should return the file to the browser as response
```
rbatistadev, line 28: Remove this comment please

web/rest/user/src/Controller/RecordedFilesZipAction.php
```diff
@@ -0,0 +1,44 @@
...
+                'application/zip'
+            );
+
+        $response->setContentDisposition(
+            ResponseHeaderBag::DISPOSITION_ATTACHMENT,
+            'recordedFile'
+        );
+
+        return $response;
+    }
```
rbatistadev, line 43: Why exactly the same two files?

web/rest/user/config/services.yaml
```diff
@@ -55,3 +55,7 @@ services:
         resource: '../src/Controller/My/*'
         tags: ['controller.service_arguments']
 
+    Controller\Provider\:
+        resource: '../src/Controller/Provider/*'
+        tags: [ 'controller.service_arguments' ]
```
rbatistadev, line 60: Is it mandatory to create a new folder for one controller?

web/rest/user/config/api/raw/provider.yml
danigargar: That name was not my choice: it's already called so in client portal and IMHO it should be named the same in both sites.

https://github.com/irontec/ivozprovider/blob/fb642fab38f65342d3b9bf6fb9dd863f5961fcb6/web/rest/client/config/api/raw/provider.yml#L1754

web/rest/user/config/services.yaml
danigargar: I'm trying to align with the current status: this file also exists in the client portal, and I'd prefer to have both follow the same folder structure.

Additionally, I see this controller as the first one, not the only one ;)

web/rest/user/src/Controller/RecordedFilesZipAction.php
danigargar: Change of mind + oversight. Deleted.

rbatistadev (approved):

ironArt3mis (approved):

PR #2740, opened 2024-08-27
Impersonation with webportal

danigargar:
Added logic for impersonating with the first suitable WebPortal

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


ironArt3mis (changes requested):
Functional review required

web/portal/platform/src/entities/Administrator/ForeignKeyResolver.tsx
```diff
@@ -15,9 +18,31 @@ const foreignKeyResolver: foreignKeyResolverType = async function ({
...
+    })
+  );
   await Promise.all(promises);
 
-  return data;
+  return data.map((row) => {
+    row.impersonationUrl = webPortals[0].url;
+
+    return row;
+  });
```
mmadariaga, line 45: Typescript types in ForeingKeyResolvers are wrong in general (red underline in the IDE). It should be something like this:
```
   const getAction = store.getActions().api.get;
-  const webPortals: Record<number, WebPortalPropertyList<unknown>> = [];
+
+  type WebPortalCollectionType = Array<Partial<WebPortalPropertyList<EntityValue>>>;
+  const webPortals: WebPortalCollectionType = [];
 
   promises.push(
     getAction({
@@ -32,13 +35,15 @@ const foreignKeyResolver: foreignKeyResolverType = async function ({
       },
       cancelToken: cancelToken,
       successCallback: async (response) => {
-        webPortals.push(...response);
+        webPortals.push(...(response as WebPortalCollectionType));
       },
     })
   );
+
   await Promise.all(promises);
 
-  return data.map((row) => {
+  type AdminCollectionType = Array<Partial<AdministratorPropertyList<EntityValue>>>;
+  return (data as AdminCollectionType).map((row) => {
     row.impersonationUrl = webPortals[0].url;
 
     return row;
```

mmadariaga: Apart from that, Foreign key resolver is supposed to do just PK => string transformations. So, it's quite unexpected to find this code here.

This should probably be handled in the impersonate component itself, but as it would require a service or some code in the store... I'll leave the decision up to you

mmadariaga (changes requested):

web/portal/platform/src/entities/WebPortal/WebPortalProperties.ts
```diff
@@ -11,7 +11,9 @@ export type WebPortalPropertyList<T> = {
   id?: T;
   logo?: T;
   urlType?: T;
+  // TODO revisar tipos
```
mmadariaga, line 14: TODO

mmadariaga (changes requested):

mmadariaga (approved):

mmadariaga (changes requested):

web/portal/brand/src/entities/Company/List.tsx
```diff
@@ -0,0 +1,39 @@
...
+      params: {
+        urlType: 'admin',
+        _pagination: false,
+      },
+      cancelToken: cancelToken,
+      successCallback: async (response) => {
+        setCustomData(response);
+      },
+    });
+  }, [rows.length]);
```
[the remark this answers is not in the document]
danigargar: row changes twice: once when the page is loaded and then again when foreign keys are resolved. If useEffect is triggered by full row the api call gets called twice.

Every thing works fine if you change pages. I'm talking about pagination.

mmadariaga (approved):

ironArt3mis (approved):

PR #2762, opened 2024-09-09
Validating the email field in the voicemail form

R0MANDEV:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description
Validating the email field in the voicemail form

### DIFF
--- a/library/Ivoz/Provider/Domain/Model/Voicemail/Voicemail.php
+++ b/library/Ivoz/Provider/Domain/Model/Voicemail/Voicemail.php
@@ -2,6 +2,7 @@
 
 namespace Ivoz\Provider\Domain\Model\Voicemail;
 
+use Ivoz\Core\Domain\Assert\Assertion;
 use Ivoz\Provider\Domain\Model\Language\LanguageInterface;
 
 /**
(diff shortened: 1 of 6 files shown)


ironArt3mis (changes requested):
Functional review required

library/Ivoz/Ast/Domain/Model/Voicemail/VoicemailAbstract.php
```diff
@@ -541,7 +541,23 @@ public function getAlias(): ?string
...
+            Assertion::true(
+                !preg_match('/\s/', $email),
+                'Email address should not contain spaces.'
+            );
+
+            Assertion::regex(
+                $email,
+                '/^[^\s@]+@[^\s@]+\.[^\s@]+$/',
+                'Invalid email address format.'
+            );
```
danigargar, line 560: `Assertion::email($email, 'Invalid email address format')`

Validating emails is tricky. In my opinion, it's better to rely on a library for this. The assertion mentioned uses PHP's built-in functionality.

If we don't need to strictly enforce email format but simply want to prevent clients from providing more than one address, we could count the occurrences of the '@' symbol and accept only one.

danigargar:
Tests not passing

danigargar (changes requested):

PR #2803, opened 2024-10-31
Avoid deleting default media relay set

danigargar:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/library/Ivoz/Provider/Domain/Service/MediaRelaySet/MediaRelaySetLifecycleServiceCollection.php
+++ b/library/Ivoz/Provider/Domain/Service/MediaRelaySet/MediaRelaySetLifecycleServiceCollection.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Ivoz\Provider\Domain\Service\MediaRelaySet;
+
+use Ivoz\Core\Domain\Assert\Assertion;
+use Ivoz\Core\Domain\Service\DomainEventSubscriberInterface;
+use Ivoz\Core\Domain\Service\LifecycleEventHandlerInterface;
+use Ivoz\Core\Domain\Service\LifecycleServiceCollectionInterface;
+use Ivoz\Core\Domain\Service\LifecycleServiceCollectionTrait;
+
+/**
+ * @codeCoverageIgnore
+ */
+class MediaRelaySetLifecycleServiceCollection implements LifecycleServiceCollectionInterface
+{
+    use LifecycleServiceCollectionTrait;
+
+    /** @var array<array-key, array> $bindedBaseServices */
+    public static $bindedBaseServices = [
+        "pre_remove" =>
+        [
+            \Ivoz\Provider\Domain\Service\MediaRelaySet\DeleteProtection::class => 200,
+        ],
+    ];
+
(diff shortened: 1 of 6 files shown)


library/Ivoz/Provider/Domain/Service/MediaRelaySet/MediaRelaySetHandlerInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Ivoz\Provider\Domain\Service\MediaRelaySet;
+
+use Ivoz\Core\Domain\Service\LifecycleEventHandlerInterface;
+use Ivoz\Provider\Domain\Model\MediaRelaySet\MediaRelaySetInterface;
+
+interface MediaRelaySetHandlerInterface extends LifecycleEventHandlerInterface
```
rbatistadev, line 8:
```suggestion
interface MediaRelaySetEventHandlerInterface extends LifecycleEventHandlerInterface
```

library/Ivoz/Provider/Domain/Service/MediaRelaySet/MediaRelaySetLifecycleServiceCollection.php
```diff
@@ -0,0 +1,31 @@
...
+    public static $bindedBaseServices = [
+        "pre_remove" =>
+        [
+            \Ivoz\Provider\Domain\Service\MediaRelaySet\DeleteProtection::class => 200,
+        ],
+    ];
+
+    protected function addService(string $event, LifecycleEventHandlerInterface|DomainEventSubscriberInterface $service): void
+    {
+        Assertion::isInstanceOf($service, MediaRelaySetHandlerInterface::class);
```
rbatistadev, line 28:
```suggestion
        Assertion::isInstanceOf($service, MediaRelaySetEventHandlerInterface::class);
```

web/rest/platform/features/provider/mediaRelaySet/removeMediaRelaySet.feature
```diff
@@ -10,3 +10,19 @@ Feature: Manage media relay sets
...
+      And I send a "DELETE" request to "/media_relay_sets/0"
+     Then the response status code should be 403
+      And the response should be in JSON
+      And the header "Content-Type" should be equal to "application/problem+json; charset=utf-8"
+      And the JSON should be like:
+      """
+      {
+        "detail": "Rejected request during security check"
+      }
+      """
```
rbatistadev, line 28: I think its enough with the status code 403, this error comes from library I think, that's the reason you use should be like and not should be equal, but its up to you

danigargar: As a general rule I want to match both the failure code and the detail. The code is not enough because it can result in a test that passes because it fails as a result of another check: for instance, because the default media relay set is being used.

library/Ivoz/Provider/Domain/Service/MediaRelaySet/MediaRelaySetLifecycleServiceCollection.php
danigargar: Thanks!

PR #2826, opened 2024-11-25
Refactorize and test microservice/recordings

danigargar:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/microservices/recordings/src/Model/RawRecordingInfo.php
+++ b/microservices/recordings/src/Model/RawRecordingInfo.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Model;
+
+class RawRecordingInfo
+{
+    protected string $fileName;
+    protected string $hashid;
+
+    public function __construct(
+        protected string $fullFileName,
+        protected string $callid,
+        protected int $size,
+        protected int $age,
+    ) {
+        $this->fileName = basename($fullFileName);
+        $this->hashid = substr(md5($callid), 0, 8);
+    }
+
+    public function getFullName(): string
+    {
+        return $this->fullFileName;
+    }
+
+    public function getFileName(): string
+    {
+        return $this->fileName;
+    }
+
+    public function getHashid(): string
+    {
+        return $this->hashid;
+    }
+
+    public function getCallid(): string
+    {
+        return $this->callid;
+    }
+
+    public function getSize(): int
+    {
+        return $this->size;
+    }
+
+    public function getAge(): int
+    {
+        return $this->age;
+    }
+}
--- a/microservices/recordings/src/Service/Encoder.php
+++ b/microservices/recordings/src/Service/Encoder.php
@@ -7,11 +7,10 @@ use Ivoz\Provider\Domain\Model\BillableCall\BillableCallInterface;
 use Ivoz\Provider\Domain\Model\BillableCall\BillableCallRepository;
 use Ivoz\Provider\Domain\Model\Recording\RecordingDto;
 use Ivoz\Provider\Domain\Model\Recording\RecordingInterface;
-use Ivoz\Provider\Domain\Model\User\UserInterface;
 use Ivoz\Provider\Domain\Model\UsersCdr\UsersCdrInterface;
 use Ivoz\Provider\Domain\Model\UsersCdr\UsersCdrRepository;
+use Model\RawRecordingInfo;
 use Symfony\Bridge\Monolog\Logger;
-use Symfony\Component\Process\Process;
 
 class Encoder
 {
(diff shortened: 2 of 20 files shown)


ironArt3mis (changes requested):
Functional review required

microservices/recordings/src/Model/RawRecordingInfo.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function getHashid(): ?string
+    {
+        return $this->hashid;
+    }
+
+    public function getCallid(): ?string
+    {
+        return ($this->callid);
```
Kaian, line 40: Remove extra ()

microservices/recordings/src/Service/RawRecordingsGetter.php
```diff
@@ -0,0 +1,44 @@
...
+            return [];
+        }
+
+        while (false !== ($filename = readdir($dir))) {
+            $absoluteFilename = sprintf(
+                "%s%s",
+                $this->rawRecordingsDir,
+                $filename
+            );
+
```
Kaian, line 36: I would add a check to only consider actual files and skip directories, links, and so on. The rawRecordingsDir may contain other files that are not actually Recordings.

Also, considering only files that ends with -mix.wav could be a good idea, because other files in the directory that doesn't match that pattern should be ingored, and not considered as errors.

microservices/recordings/tests/Service/EncoderTest.php
```diff
@@ -0,0 +1,200 @@
...
+            $this->rawRecordingInfoMockFactory(
+                '/recordings/c00269fa-a64b-8297bdde-309cd49f@10.10.1.125-fc5cee56dc74-mix.wav',
+                Encoder::RECORDING_SIZE_MIN + 1,
+                Encoder::RECORDING_AGE_MIN + 1
+            ),
+            $this->rawRecordingInfoMockFactory(
+                '/recordings/wrong-file-name.ogg',
+                Encoder::RECORDING_SIZE_MIN + 1,
+                Encoder::RECORDING_AGE_MIN + 1
+            ),
```
Kaian, line 131: This file should be ignored by the recording microservice

microservices/recordings/src/Model/RawRecordingInfo.php
```diff
@@ -0,0 +1,52 @@
...
+    public function __construct(
+        protected string $file,
+    ) {
+
+        $this->filename = basename($this->file);
+
+        if (preg_match("/\w+-\w+-(.*)-\w+-mix.wav/", $this->filename, $matches)) {
+            $this->callid = urldecode($matches[1]);
+            $this->hashid = substr(md5($this->callid), 0, 8);
+        }
```
Kaian, line 20: I will validate this before creating a RawRecordingInfo instance. If this condition doesn't match, it's not a recording and doesn't deserve a RawRecordingInfo. Also $this->callid and $this->hashid shouldn't be nullable. If the fille doesn't have those values, it can not be considered a recording, so callid could be part of the constructor imo.

microservices/recordings/src/Service/Encoder.php
```diff
@@ -25,292 +24,381 @@ class Encoder
...
+
+    /**
+     * @return UsersCdrInterface[]|BillableCallInterface[]
+     */
+    private function getAccCdrs(RawRecordingInfo $filename): array
+    {
+        $billableCalls = $this
+            ->billableCallRepository
+            ->findByCallid(
+                $filename->getCallid() ?? ''
```
Kaian, line 318: As mentioned before, this should not happen. Callid is a must in the recording and should not be nullable.

Kaian (changes requested):

microservices/recordings/src/Service/RawRecordingsGetter.php
Kaian: Maybe this is also a good point to validate file format to extract callid from the filename

danigargar: This service is here only keep file system functions separated from the logic. It's fully mocked in testing.

microservices/recordings/tests/Service/EncoderTest.php
danigargar: And it is ignored: 
"[Recordings][ERROR] Unable to get info from filename wrong-file-name.ogg\n",

This log line above is required for tests to pass.

Kaian: I mean it should not be considered at all by the logic of the encoder, not even in logs. At least it was with the original code iirc. You could assume the directory of raw recording files can contain some files that has to be parsed (that have an specific format) and thousands that does not require parsing at all.

danigargar: Just checked it in bleeding and that error message was already there.

Kaian: Is a surprise to me that the bleeding code parses files that doesn't have a raw rtpengine recording format, it should at least end with -mix.wav or something like that :thinking:

microservices/recordings/src/Model/RawRecordingInfo.php
danigargar: RawRecordingInfo objects are partially mocked in testing. I could move this file name logic and this object creation to Encoder, but then I won‘t be able to mock them.

microservices/recordings/src/Model/RawRecordingInfo.php
```diff
@@ -4,49 +4,46 @@
...
+        protected int    $age,
     ) {
-
-        $this->filename = basename($this->file);
-
-        if (preg_match("/\w+-\w+-(.*)-\w+-mix.wav/", $this->filename, $matches)) {
-            $this->callid = urldecode($matches[1]);
-            $this->hashid = substr(md5($this->callid), 0, 8);
-        }
+        $this->fileName = basename($this->fullFileName);
```
Kaian, line 16: Based on the line below of how to calculate $hashid: 

      $this->fileName = basename($fullFileName);

microservices/recordings/src/Service/Encoder.php
```diff
@@ -275,7 +275,7 @@ private function getRecordingFromUserCdr(UsersCdrInterface $usersCdr): ?Recordin
 
     private function getHashId(?string $callid): string
```
Kaian, line 276: private function getHashId(string $callid): string

microservices/recordings/src/Service/RawRecordingInfoFactory.php
```diff
@@ -0,0 +1,36 @@
...
+
+use Model\RawRecordingInfo;
+
+class RawRecordingInfoFactory
+{
+    public function createRawRecordingInfo(string $fullFileName): ?RawRecordingInfo
+    {
+        $fileName = basename($fullFileName);
+
+        if (preg_match("/\w+-\w+-(.*)-\w+-mix.wav/", $fileName, $matches)) {
```
Kaian, line 13: Other services have regexp patterns in constants, but up to you :+1:

microservices/recordings/tests/Service/EncoderTest.php
```diff
@@ -107,35 +143,40 @@ private function rawRecordingInfoMockFactory(string $file, int $size, int $age):
...
                 Encoder::RECORDING_SIZE_MIN + 1,
                 Encoder::RECORDING_AGE_MIN + 1
             ),
-            $this->rawRecordingInfoMockFactory(
+//            $this->rawRecordingInfoMockFactory(
+//                '/recordings/wrong-file-name-mix.wav',
+//                ''
+//                Encoder::RECORDING_SIZE_MIN + 1,
+//                Encoder::RECORDING_AGE_MIN + 1
+//            ),
```
Kaian, line 170: Commented code here

microservices/recordings/src/Service/Encoder.php
```diff
@@ -25,292 +24,372 @@ class Encoder
...
+                ->execute($convertWav, $metadata, $convertMp3);
+        } catch (\Exception $e) {
+            $this->stats['error']++;
+            $this->logger->info(sprintf(
+                "[Recordings][%s] Failed to convert audio: %s\n",
+                $rawRecordingInfo->getHashid(),
+                $e->getMessage()
+            ));
+
+            return 0;
```
Kaian, line 161: This function always returns 0, but is not the duration of the mp3 the expected return value?

Kaian (changes requested):

microservices/recordings/src/Service/Encoder.php
Kaian: My bad, didn't check the diff properly

Kaian (approved):

ironArt3mis (approved):

cruzccl (approved):

PR #2848, opened 2024-12-30
Extensions selector do not works fine

rbatistadev:
#### Type Of Change 
- [X] Small bug fix
- [ ] Breaking change

#### Checklist:
- [X] Commits are split per component (schema, portal/platform, kamusers, agis, ..)


web/portal/client/src/entities/Extension/Extension.tsx
```diff
@@ -109,7 +109,9 @@ const properties: ExtensionProperties = {
   },
   user: {
     label: _('User', { count: 1 }),
-    required: true,
+    required: false,
+    default: '__null__',
+    null: _('Unassigned'),
```
danigargar, line 114: I have some doubts about this change: it's not requested but it makes sense if you can not possibly get a user extension with null user.

Nothing wrong with your code, but this may suggest a definition error.

danigargar (changes requested):
Some tests are failing

web/portal/client/src/entities/Extension/Extension.tsx
rbatistadev: In database definition, you can place a user extension but not a user, user is a nullable field with a null default value

danigargar: I know that. My concerns are:
·  Should an extension with unassigned type be assignable to an user?
· You can actually add extensions of type user without user using the mass importer.

I'm just pointing the issue may need some clarification. 
I'll approve it as soon as all test are green.

ironArt3mis (changes requested):
Functional review required

danigargar (approved):

danigargar (approved):

danigargar (changes requested):

danigargar (approved):

ironArt3mis (approved):

PR #2918, opened 2025-03-10
[PROVIDER-2202] New survival device entity

R0MANDEV:
#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:


ironArt3mis (changes requested):
Functional review required

web/rest/client/features/provider/survivalDevice/putSurvivalDevice.feature
```diff
@@ -0,0 +1,42 @@
...
+      {
+        "name": "Survival Test 1-1",
+        "proxy": "55555",
+        "outboundProxy": "66666",
+        "udpPort": 5065,
+        "tcpPort": 5065,
+        "tlsPort": 5066,
+        "wssPort": 10085,
+        "description": "new survival device 1-1",
+        "company": 1
```
danigargar, line 22: Must work without _company_, so it's better to remove this line.

web/rest/client/public/swagger.json
danigargar, line 1: Why did you add this file?

Remove it or explain why we should add it.

web/portal/client/src/entities/SurvivalDevice/SurvivalDevice.tsx
```diff
@@ -0,0 +1,63 @@
...
+    default: 0,
+  },
+  tlsPort: {
+    label: _('TLS port'),
+    default: 0,
+  },
+  wssPort: {
+    label: _('WSS port'),
+    default: 0,
+  },
```
danigargar, line 37: It would be better to set default values to match the schema defaults.

web/portal/client/src/entities/SurvivalDevice/SurvivalDevice.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const columns = ['name', 'proxy', 'description'];
+
+const survivalDevice: EntityInterface = {
+  ...defaultEntityBehavior,
+  icon: WatchLaterIcon,
+  iden: 'SurvivalDevice',
+  title: _('Survival Devices', { count: 2 }),
```
danigargar, line 46: `title: _('Survival Device', { count: 2 }),`

danigargar (changes requested):

danigargar (changes requested):
##### Commit 82135a4 message is:
`core: Improved entity generators to add extra DTO methods`
It have only fixture changes or aditions, so tag it accordingly. Something like:
`tests: updated fixtures` would work.

web/portal/client/src/translations/it.json
```diff
@@ -449,6 +450,7 @@
   "Priority": "Priorità",
   "Private": "Private",
   "Provisioning Info": "Informazioni di provisioning",
+  "Proxy": "Procuratore",
```
danigargar, line 453: I seriously doubt that they say "procuratore" for proxy in this context.
Leave it as Proxy, please

web/rest/client/features/provider/survivalDevice/putSurvivalDevice.feature
danigargar: Please, remove company!

danigargar (changes requested):

danigargar (approved):

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+            throw new \DomainException("Invalid port number: $port. Must be between 1 and 65535.");
+        }
+
+        return $port;
+    }
+
+    public function setUdpPort(int $udpPort): static
+    {
+        return parent::setUdpPort($this->sanitizePort($udpPort));
+    }
```
danigargar, line 53:
```suggestion
    public function setUdpPort(int $udpPort): static
    {
        Assertion::between($port,1, 65535, "Invalid udp port number: $udpPort. Must be between 1 and 65535");
        return parent::setUdpPort($this->sanitizePort($udpPort));
    }
```

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function setTcpPort(int $tcpPort): static
+    {
+        return parent::setTcpPort($this->sanitizePort($tcpPort));
+    }
```
danigargar, line 58:
```suggestion
    public function setTcpPort(int $tcpPort): static
    {
         Assertion::between($port,1, 65535, "Invalid tcp port number: $tcpPort. Must be between 1 and 65535");
        return parent::setTcpPort($this->sanitizePort($tcpPort));
    }
```

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function setTlsPort(int $tlsPort): static
+    {
+        return parent::setTlsPort($this->sanitizePort($tlsPort));
+    }
```
danigargar, line 63:
```suggestion
    public function setTlsPort(int $tlsPort): static
    {
               Assertion::between($port,1, 65535, "Invalid tls port number: $tlsPort. Must be between 1 and 65535");
        return parent::setTlsPort($this->sanitizePort($tlsPort));
    }
```

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function setWssPort(int $wssPort): static
+    {
+        return parent::setWssPort($this->sanitizePort($wssPort));
+    }
```
danigargar, line 68:
```suggestion
    public function setWssPort(int $wssPort): static
    {
               Assertion::between($port,1, 65535, "Invalid wss port number: $wssPort. Must be between 1 and 65535");
        return parent::setWssPort($this->sanitizePort($wssPort));
    }
```

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+            }
+        } else {
+            $this->setOutboundProxy(null);
+        }
+        return parent::setOutboundProxy($outboundProxy);
+    }
+
+    protected function setProxy(?string $proxy): static
+    {
+        if (! $proxy) {
```
danigargar, line 88:
```suggestion
        if (!$proxy) {
```

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+    protected function sanitizeValues(): void
+    {
+        $this->udpPort = $this->sanitizePort($this->udpPort);
+        $this->tcpPort = $this->sanitizePort($this->tcpPort);
+        $this->tlsPort = $this->sanitizePort($this->tlsPort);
+        $this->wssPort = $this->sanitizePort($this->wssPort);
+
+        $this->setProxy($this->proxy);
+        $this->setOutboundProxy($this->outboundProxy);
+    }
```
danigargar, line 39: Remove this method.

- _sanitizeValues_ is used to validate the entity as a whole: checks that look more than one field at a time.
- You should access values through getters.
- Don't call setters here: that only sets the value it already has.
- Port assignations are are useless here.

Anyway, is better to validate each port separately on each setter. Look below.

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,99 @@
...
+            $outboundProxy = trim($outboundProxy);
+            if ($outboundProxy === '') {
+                throw new \DomainException('Outbound Proxy cannot be empty.');
+            }
+            if ($this->getProxy() && strpos($this->getProxy(), ':') !== false) {
+                throw new \DomainException('Proxy must not include a port when Outbound Proxy is set.', 70003);
+            }
+        } else {
+            $this->setOutboundProxy(null);
+        }
```
danigargar, line 82:
```suggestion
        if ($this->getOutboundProxy()) {
            $outbound_proxy = explode(':', $this->getOutboundProxy());
            $ip = array_shift($outbound_proxy);
            $obPort = array_shift($outbound_proxy);
            if (!is_null($port)) {
                throw new \DomainException('When Outbound Proxy is used, SIP Proxy must not include a port.', 70003);
            }
        }    
```

Or am I missing something?

Anyway you have an infinite recursive loop if _$outboundProxy_ is _null_.

danigargar (changes requested):

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,90 @@
...
+            throw new \DomainException('Proxy cannot be null');
+        }
+
+        $proxy = trim($proxy);
+        if ($proxy === "") {
+            throw new \DomainException('Proxy cannot be empty');
+        }
+
+        return parent::setProxy($proxy);
+    }
```
danigargar, line 89:
```suggestion
    protected function setProxy(string $proxy): static
    {
        $proxy = trim($proxy);
        if (empty($proxy) {
            throw new \DomainException('Proxy cannot be empty');
        }

        return parent::setProxy($proxy);
    }
```

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,98 @@
...
+
+    private function sanitizePort(int $port): int
+    {
+        if ($port < 1 || $port > 65535) {
+            throw new \DomainException("Invalid port number: $port. Must be between 1 and 65535.");
+        }
+
+        return $port;
+    }
```
danigargar, line 97: No need to do this.
You already are making an Assertion before each call to this method, so this exception will never be thrown.

Remove this method.

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,98 @@
...
+                throw new \DomainException('Outbound Proxy cannot be empty.');
+            }
+
+            if ($this->getProxy() && strpos($this->getProxy(), ':') !== false) {
+                throw new \DomainException('Proxy must not include a port when Outbound Proxy is set.', 70003);
+            }
+        } else {
+            $this->setOutboundProxy(null);
+        }
+    }
```
danigargar, line 88:
```suggestion
    protected function sanitizeProxyLogic(): void
    {
        $outboundProxy = $this->getOutboundProxy();

        if (is_null($outboundProxy)) {
            return;
        }
        
        $emptyOutboundProxy = empty(
            trim($outboundProxy)
        );

        if ($emptyOutboundProxy) {
            throw new \DomainException('Outbound Proxy cannot be empty.');
        }

        if ($this->getProxy() && strpos($this->getProxy(), ':') !== false) {
            throw new \DomainException('Proxy must not include a port when Outbound Proxy is set.', 70003);
        }
    }
```

- Choose guards (condition -> return) over if/else blocks
- _$this->setOutboundProxy(null);_ only gets called if OutboundProxy is already null.

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->sanitizeProxyLogic();
+    }
+
+    public function setUdpPort(int $udpPort): static
+    {
+        Assertion::between($udpPort, 1, 65535, "Invalid udp port number: $udpPort. Must be between 1 and 65535");
+        return parent::setUdpPort($this->sanitizePort($udpPort));
+    }
+
+    public function setTcpPort(int $tcpPort): static
```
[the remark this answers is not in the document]
R0MANDEV: listo

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        Assertion::between($tcpPort, 1, 65535, "Invalid tcp port number: $tcpPort. Must be between 1 and 65535");
+        return parent::setTcpPort($this->sanitizePort($tcpPort));
+    }
+
+    public function setTlsPort(int $tlsPort): static
```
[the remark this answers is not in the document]
R0MANDEV: listo

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        Assertion::between($tlsPort, 1, 65535, "Invalid tls port number: $tlsPort. Must be between 1 and 65535");
+        return parent::setTlsPort($this->sanitizePort($tlsPort));
+    }
+
+    public function setWssPort(int $wssPort): static
```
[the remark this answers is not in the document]
R0MANDEV: listo

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return $this->id;
+    }
+
+    protected function sanitizeValues(): void
+    {
+        $this->sanitizeProxyLogic();
+    }
+
+    public function setUdpPort(int $udpPort): static
```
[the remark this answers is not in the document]
R0MANDEV: listo

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
R0MANDEV: listo

library/Ivoz/Provider/Domain/Model/SurvivalDevice/SurvivalDevice.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $emptyOutboundProxy = empty(
+            trim($outboundProxy)
+        );
+        if ($emptyOutboundProxy) {
+            throw new \DomainException('Outbound Proxy cannot be empty.');
+        }
+        if ($this->getProxy() && strpos($this->getProxy(), ':') !== false) {
+            throw new \DomainException('Proxy must not include a port when Outbound Proxy is set.', 70003);
+        }
```
danigargar, line 86:
```suggestion
        $sipProxyIncludesPort = strpos($this->getProxy(), ':') !== false;

        if ($sipProxyIncludesPort) {
            throw new \DomainException('When Outbound Proxy is used, SIP Proxy must not include a port.', 70003);
        }                
```

1. We usually define variables to clarify our intentions. In this case "$sipProxy includes port" for we are checking whether proxy includes a port or not. And out condition know can read almost like in plain english "if sip proxy includes port..."
2. There's no need to check whether "getProxy()" returns truthy. It's always truthy if executions reaches this point, and implies forcing proxy to be non-empty out of place and with the wrong exception.
3. We should be using exactly the same exception message as in CarrierServer unless told otherwise.

danigargar (changes requested):

danigargar:
> al colocar protected a setUdpPort, setTcpPort, setTlsPort, setWssPort, da errores debe ser public porque la interfaz SurvivalDeviceInterface lo define como public.

In english, please!

The issue was the entitiy generator created the interface signatures matching those on the SurvivaDevice: if you make a method public and run the generator, you'll get a signature for at the interface

danigargar (approved):

ironArt3mis (approved):

PR #2985, opened 2025-07-02
[PROVIDER-2285] Null price fields if showInvoices is false

R0MANDEV:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:


web/rest/client/src/Service/Behat/FeatureContext.php
danigargar, line 1: What's the issue here. We already have a "Should be like matcher".
And if this was really necessary it should added on api.

danigargar (changes requested):

danigargar (approved):

PR #3020, opened 2025-09-22
Fix leading spaces in provision templates

R0MANDEV:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:


library/Ivoz/Provider/Domain/Service/TerminalModel/TemplateRenderer.php
```diff
@@ -100,8 +100,8 @@ public function run()
                 {
                     error_reporting(error_reporting() & ~E_NOTICE);
                     ?>
-                    [__TEMPLATE_BODY__]
-                    <?php
+[__TEMPLATE_BODY__]
+<?php
```
danigargar, line 104:
```suggestion
                   ?>[__TEMPLATE_BODY__]<?php
```

This should produce the same effect. You may need a couple `echo PHP_EOL` if you miss some new line before/after. 

Sorry, but that indent hurts.

PR #3034, opened 2025-10-31
[PROVIDER-2361] fix specificUrlPattern logic

cruzccl:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

#####  Problem

  When provisioning terminals via HTTPS, the system was executing specific templates regardless of the requested file extension. For example, requesting aabbccddeeff.config would execute
   the template as if aabbccddeeff.cfg was requested, even when the specificUrlPattern was set to {mac}.cfg.

#####  Root Cause

  In ProvisionSpecific::getTerminalByUrl(), the extension validation was comparing the requested URL extension against itself instead of comparing it against the extension defined in the
   terminal model's specificUrlPattern.

#####  New Behavior

  • If specificUrlPattern is {mac}.cfg:
    • ✅ aabbccddeeff.cfg → Template executed
    • ❌ aabbccddeeff.other → 404 (terminal not found)
  • If specificUrlPattern is {mac} (no extension):
    • ✅ Any extension is accepted (expected behavior)

### DIFF
--- a/library/Ivoz/Provider/Domain/Service/Terminal/Provision/ProvisionSpecific.php
+++ b/library/Ivoz/Provider/Domain/Service/Terminal/Provision/ProvisionSpecific.php
@@ -152,7 +152,7 @@ class ProvisionSpecific
                 $terminalModel->getSpecificUrlPattern()
             );
             $specificUrlExtension = $this->extractFileExtension(
-                $specificUrlPattern
+                $terminalModel->getSpecificUrlPattern()
             );
             $fixedUrlSegments = explode('{mac}', strtolower($specificUrl), 2);
             $fixedSpecificUrl = str_ireplace('{mac}', (string) $candidate->getMac(), $specificUrl);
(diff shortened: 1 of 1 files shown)


library/Ivoz/Provider/Domain/Service/Terminal/Provision/ProvisionSpecific.php
```diff
@@ -191,8 +191,12 @@ private function getTerminalByUrl(string $specificUrlPattern): ?TerminalInterfac
         return null;
     }
 
-    protected function extractFileExtension(string $route): string
+    protected function extractFileExtension(string|null $route): string
```
danigargar:
```suggestion
    protected function extractFileExtension(?string $route): string
```

Sorry, we always use this syntax for nullable params.

danigargar (changes requested):

danigargar (approved):

PR #3035, opened 2025-11-03
[PROVIDER-2350] Filter carriers without servers from outgoing routing selection

R0MANDEV:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/web/portal/brand/cypress/fixtures/Provider/BillableCalls/getItem.json
+++ b/web/portal/brand/cypress/fixtures/Provider/BillableCalls/getItem.json
@@ -15,14 +15,17 @@
     "endpointName": null,
     "direction": "outbound",
     "id": 1,
-    "company": "~",
+    "company": {
+      "id": 1,
+      "name": "Company 1"
+    },
     "carrier": "~",
     "destination": null,
     "ratingPlanGroup": null,
     "invoice": "~",
     "ddi": "~",
     "ddiProvider": "~"
-},
+  },
   "headers": {
     "content-type": "application/json; charset=utf-8"
   },
--- a/web/portal/brand/cypress/pacts/brand-provider-External-calls-brand-consumer-External-calls.json
+++ b/web/portal/brand/cypress/pacts/brand-provider-External-calls-brand-consumer-External-calls.json
@@ -40,7 +40,10 @@
           "endpointName": null,
           "direction": "outbound",
           "id": 1,
-          "company": "~",
+          "company": {
+            "id": 1,
+            "name": "Company 1"
+          },
           "carrier": "~",
           "destination": null,
           "ratingPlanGroup": null,
--- a/web/portal/brand/cypress/pacts/brand-provider-ddis-billable_calls-brand-consumer-ddis-billable_calls.json
+++ b/web/portal/brand/cypress/pacts/brand-provider-ddis-billable_calls-brand-consumer-ddis-billable_calls.json
@@ -40,7 +40,10 @@
           "endpointName": null,
           "direction": "outbound",
           "id": 1,
-          "company": "~",
+          "company": {
+            "id": 1,
+            "name": "Company 1"
+          },
           "carrier": "~",
           "destination": null,
           "ratingPlanGroup": null,


ironArt3mis (changes requested):

library/Ivoz/Provider/Domain/Model/Carrier/CarrierRepository.php
```diff
@@ -45,4 +45,14 @@ public function countByBrand(int $brandId): int;
...
+
+    /**
+     * @return CarrierInterface[]
+     */
+    public function findForLcrRoutesForBrand(int $brandId): array;
+
+    /**
+     * @return CarrierInterface[]
+     */
+    public function findForStaticRoutesForBrand(int $brandId): array;
```
danigargar:
```suggestion
    public function findStaticRoutesForBrand(int $brandId): array;
```

library/Ivoz/Provider/Domain/Service/OutgoingRouting/CheckCarrierHasServers.php
```diff
@@ -0,0 +1,43 @@
...
+        if (!$carrier) {
+            return;
+        }
+
+        $servers = $carrier->getServers();
+        if (count($servers) === 0) {
+            throw new \DomainException(
+                'Carrier has no CarrierServers'
+            );
+        }
```
danigargar: Do you see any impediment for checking this at OutgoingRouting.setCarrier?

```php
protected function setCarrier(CarrierInterface $carrier): static
{
          if ($this->getRoutingMode() === OutgoingRoutingInterface::ROUTINGMODE_BLOCK) {
            return;
        } 

        $hasServers = !empty($carrier->getServers());

        if ($hasServer) {
             return;
        }

         throw new \DomainException('Carrier has no servers');
}
```

library/Ivoz/Provider/Domain/Service/OutgoingRoutingRelCarrier/CheckCarrierHasServers.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $carrier = $relCarrier->getCarrier();
+
+        $servers = $carrier->getServers();
+        if (count($servers) === 0) {
+            throw new \DomainException(
+                'Carrier has no CarrierServers'
+            );
+        }
+    }
```
danigargar: You could check this straight at OutgoingRoutingRelCarrier::setCarrier.

schema/tests/Provider/CarrierServer/CarrierServerDeleteProtectionTest.php
danigargar: What are this changes for?

Other than for IAs compulsion for calling entityManager, I don't see the point.

web/portal/brand/src/entities/OutgoingRouting/hooks/useCarriers.ts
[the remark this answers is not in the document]
R0MANDEV: Very good idea, I'm going to put it into practice, I think it's good.

web/portal/brand/src/entities/OutgoingRouting/Form.tsx
```diff
@@ -34,10 +34,17 @@ const Form = (props: EntityFormProps): JSX.Element => {
...
+    fkChoices = {
+      ...fkChoices,
+      carrierIds: carriers,
+    };
+  } else if (routingMode === 'static') {
+    fkChoices = {
+      ...fkChoices,
+      carrier: carriers,
+    };
+  }
```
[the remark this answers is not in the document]
R0MANDEV: You're right that carriers has the same type in both cases, but we need to assign it to both fields because:

  Before this PR:
  - Only carrierIds (lcr mode) used the filtered carriers from the hook
  - carrier (static mode) used the default options from foreignKeyGetter (all carriers, unfiltered)

  After this PR:
  - Both modes need to show only carriers with servers
  - useCarriers now filters for both modes: hasServers && (routingMode === 'static' || calculateCost)
  - So we must assign the filtered options to:
    - carrier when routingMode === 'static'
    - carrierIds when routingMode === 'lcr'

  If we only assigned to carrierIds, the static mode would still show carriers without servers (using the
  default foreignKeyGetter), which defeats the purpose of this PR.

  The fields carrier (single select) and carrierIds (multi-select) are different properties in the model,
  controlled by visualToggle, so only one is visible at a time.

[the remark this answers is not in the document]
R0MANDEV: I simplify it
```
fkChoices = {
    ...fkChoices,
    carrier: carriers,
    carrierIds: carriers,
  };

```

danigargar (changes requested):

PR #3039, opened 2025-11-13
[PROVIDER-2366] Add company type validation for RetailAccount, ResidentialDevice, and Friend entities

R0MANDEV:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

### DIFF
--- a/web/rest/client/features/my/dashboard/getDashboard.feature
+++ b/web/rest/client/features/my/dashboard/getDashboard.feature
@@ -78,6 +78,26 @@ Feature: Retrieve dashboard
                   "description": ""
               },
               {
+                  "name": "testRetailAccount5",
+                  "outgoingDdi": "",
+                  "description": ""
+              },
+              {
+                  "name": "testRetailAccount4",
+                  "outgoingDdi": "",
+                  "description": ""
+              },
+              {
+                  "name": "testRetailAccount3",
+                  "outgoingDdi": "",
+                  "description": ""
+              },
+              {
+                  "name": "testRetailAccount2",
+                  "outgoingDdi": "",
+                  "description": ""
+              },
+              {
                   "name": "testRetailAccount",
                   "outgoingDdi": "",
                   "description": ""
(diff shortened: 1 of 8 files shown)


library/Ivoz/Provider/Domain/Service/Friend/CheckCompanyType.php
```diff
@@ -0,0 +1,41 @@
...
+            throw new \DomainException(
+                sprintf(
+                    'Friend can only be associated with vpbx companies. Current company type: %s',
+                    $company->getType()
+                ),
+                30009
+            );
+        }
+    }
+}
```
danigargar: This check can be easily done in Friend's company setter:

```php
    protected function setCompany(CompanyInterface $company): static
    {
        if ($company->getType() !== CompanyInterface::TYPE_VPBX) {
            throw new \DomainException('Friend can only be associated with vpbx companies.');
        }

        return parent::setCompany($company);
    }
```

library/Ivoz/Provider/Domain/Service/ResidentialDevice/CheckCompanyType.php
```diff
@@ -0,0 +1,41 @@
...
+            self::EVENT_PRE_PERSIST => self::PRE_PERSIST_PRIORITY
+        ];
+    }
+
+    /**
+     * @param ResidentialDeviceInterface $residentialDevice
+     *
+     * @return void
+     */
+    public function execute(ResidentialDeviceInterface $residentialDevice): void
```
danigargar: That info is in the signature already.

```suggestion
    public function execute(ResidentialDeviceInterface $residentialDevice): void
```

danigargar (changes requested):
Life cycles are (relativlely) expensive. It's ok to use them when we need it, it's not that we have to avoid them at all cost.

But in this case it's straigthforward: you only have to check one field (company) in an entity, so use the setters.

danigargar (approved):

PR #3131, opened 2026-08-16
[PROVIDER-2479] kamailio: improve mediaRelaySet selection logic

cruzccl:
#### Type Of Change 
- [x] Small bug fix
- [ ] Breaking change

#### Checklist:

#### Description

Fixes how the media relay set (rtpengine set) is selected in both proxies.

**kamtrunks**
- Always call `set_rtpengine_set()`, default set included. The module does not fall back
  to the default set on its own — it keeps the selected one in per-process state and only
  resets it when `msg->id` changes — so on gateway failover a carrier using the default set
  kept using the set of the previous one.
- Inbound calls to unassigned DDIs never got a set assigned and were always relayed through
  the default one; they now honour the ddiProvider's.
- Bounced calls used the set of the carrier `SELECT_NEXT_GW` had picked, even though no
  carrier takes part in the call; they now use the company's.

**kamusers**
- Same explicit selection as above. Nothing is known to misroute here, as the set is read
  once from the caller's company; it just removes the trap.
- On demand recording (in-dialog INFO/MESSAGE) never selected the set, so
  `start_recording()`/`stop_recording()` were sent to the default one: for companies with a
  non-default set the command reached an rtpengine that does not know the call-id and
  recording never started. Selection now lives in its own `SET_MEDIA_RELAY_SET` route,
  called from both places.

**Both**: dropped the unused `setid_avp` modparam (`$avp(setid)` is never read nor written,
and declaring it suggested the set travels in an AVP when it does not). No behaviour change.

### DIFF
--- a/kamailio/trunks/config/kamailio.cfg
+++ b/kamailio/trunks/config/kamailio.cfg
@@ -1727,6 +1727,10 @@ route[RELAY] {
         $fs = "udp:" + $var(trunksAddress) + ":" + SIP_PORT;
         $dlg_var(carrierId) = $null;
         $dlg_var(calculateCost) = "0";
+
+        # No carrier involved: use the company's set instead of the one of the
+        # carrier SELECT_NEXT_GW happened to pick
+        $dlg_var(mediaRelaySetId) = $dlg_var(companyMediaRelaySetsId);
     }
 
     if ($(du{uri.host}) != $null) {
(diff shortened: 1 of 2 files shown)


ironArt3mis (changes requested):

Kaian (approved):

ironArt3mis (approved):

PR #3136, opened 2026-08-31
[PROVIDER-2486] doc: added agent instructions in AGENTS.md

R0MANDEV:
The repository had no instructions file for AI coding agents. This adds AGENTS.md (read directly by Codex and OpenCode) and a one-line CLAUDE.md that imports it (@AGENTS.md) for Claude Code, so one file serves all three. Every path, script, container and rule in it was checked against the repository:

- Monorepo map: library/Ivoz contexts, schema, the four REST APIs and portals, PHP and Go microservices, asterisk/agi, kamailio, cgrates, profiles, debian packaging, tests and docs.
- Where commands run: ivozprovider-backend (/opt/irontec/ivozprovider) and the per-app portal containers; the ../ivoz-ui sibling mount.
- Backend architecture and patterns: Domain/Application/Infrastructure per context, generated vs hand-written entity files and the run-generators / run-service-update / migrations flow, lifecycle service collections, domain events and jobs, assemblers, use cases, multitenancy in the API layer, phpstan/psalm/rector baselines.
- Frontend architecture: ivoz-ui portals, ACL-filtered EntityMap, app store slices, entity folder files measured across the four apps, i18n, Cypress + Pact and api-spec sync, lint.
- Commands: library/bin/test-*, schema/bin/*, REST and portal scripts.
- Git conventions summarised from doc/dev/en/commits.md and the PR template: PROVIDER-<ticket>-<slug> branches, tagged commits split per component, which tags gate which Jenkins stage, PR title without the ticket prefix.

#### Type Of Change 
- [ ] Small bug fix
- [ ] Breaking change

#### Checklist:
- [ ] Commits are named and have tag following [commit rules](https://github.com/irontec/ivozprovider/blob/main/doc/dev/en/commits.md)
- [ ] Commits are split per component (schema, portal/platform, kamusers, agis, ..)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,151 @@
+# AGENTS.md
+
+Instructions for AI coding agents. Codex and OpenCode read this file; Claude Code reads it through `CLAUDE.md` (`@AGENTS.md`). Edit this file, not `CLAUDE.md`.
+
+## Project
+
+IvozProvider — multitenant VoIP platform (Global admin → Brands → Companies → Users). Monorepo:
+
+- `library/` — shared PHP domain code (`Ivoz\{Provider,Ast,Kam,Cgr}`) on `ivoz-core`, used by every PHP app below
+- `schema/` — migrations, entity generation, ORM tests
+- `web/rest/{platform,brand,client,user}` — one REST API per admin level
+- `web/portal/{platform,brand,client,user}` — one React portal per API, on `ivoz-ui`
+- `microservices/` — PHP and Go services
+- `asterisk/` — FastAGI app (`agi/`) and Asterisk config
+- `kamailio/` — SIP proxy configs (`users/`, `trunks/`)
+- `cgrates/` — rating engine config
+- `profiles/` — per-host-role filesystem overlays
+- `debian/` — packaging
+- `tests/` — test images, black-box SIP scenarios, API and security tests
+- `doc/` — user docs and `dev/` conventions
+
+## Shared ivoz-* repositories and libraries
+
+- Backend, separate repositories installed as composer packages: `irontec/ivoz-core` and `ivoz-api-bundle` (DDD/API framework: generated entities, DTOs, assemblers, lifecycles), `ivoz-dev-tools` (entity generator), `prophecy`.
+- Backend, packages that live **inside this monorepo** (`library/composer-packages/irontec/`, installed via a composer `path` repository): `ivoz-provider-bundle`, `replacements`. Edit them here, in the same PR.
+- `irontec/prophecy` is a fork of phpspec's prophecy carrying a single fix and should eventually go back to mainstream: treat it as read-only, never add changes there.
+- Frontend: `@irontec/ivoz-ui` (admin CRUD framework, hygen templates, base translations).
+- ivozprovider may not be the only repository a change touches.
+- If a behaviour comes from an `ivoz-*` library, inspect the library before writing a workaround here.
+- Fix generic behaviour in the library when it benefits several projects.
+- Don't duplicate what an `ivoz-*` package already provides.
+- Library sources are often sibling checkouts (`../ivoz-core`, `../ivoz-ui`; `docker-compose.yml` mounts `../ivoz-ui/` at `/opt/ivoz-ui` in the portal containers). Check for them before a workaround; otherwise read `library/vendor/irontec/` or `web/portal/node_modules/@irontec/ivoz-ui/`.
+
+## Where commands run
+
+- PHP (library, schema, REST apps, microservices, agi): `docker exec -it ivozprovider-backend bash` → `/opt/irontec/ivozprovider`. Every `bin/test-*` script hardcodes that path.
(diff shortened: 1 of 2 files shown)


ironArt3mis (changes requested):

AGENTS.md
```diff
@@ -0,0 +1,114 @@
...
+- `kamailio/` — SIP proxy configs (`users/`, `trunks/`)
+- `cgrates/` — rating engine config
+- `profiles/` — per-host-role filesystem overlays
+- `debian/` — packaging
+- `tests/` — test images, black-box SIP scenarios, API and security tests
+- `doc/` — user docs and `dev/` conventions
+
+## Internal ivoz-* libraries
+
+- Backend: `irontec/ivoz-core`, `ivoz-api-bundle` (DDD/API framework: generated entities, DTOs, assemblers, lifecycles), `ivoz-provider-bundle`, `ivoz-dev-tools` (entity generator), `replacements`, `prophecy`.
```
danigargar: irontec/prophecy is a special case: it's a fork, our only contribution to it is a fix and this dependency should go back to mainstream at some point . We should handle this library as read-only.

ivoz-provider-bundle and replacements are internal packages. Maybe it should be said, or maybe not, because agents would notice at first sight. Not sure about that.

AGENTS.md
```diff
@@ -0,0 +1,114 @@
...
+- Fix generic behaviour in the library when it benefits several projects (isbc and aps use the same core/api/ui libraries).
+- Don't duplicate what an `ivoz-*` package already provides.
+- Library sources are often sibling checkouts (`../ivoz-core`, `../ivoz-ui`; `docker-compose.yml` mounts `../ivoz-ui/` at `/opt/ivoz-ui` in the portal containers). Check for them before a workaround; otherwise read `library/vendor/irontec/` or `web/portal/node_modules/@irontec/ivoz-ui/`.
+
+## Where commands run
+
+- PHP (library, schema, REST apps, microservices, agi): `docker exec -it ivozprovider-backend bash` → `/opt/irontec/ivozprovider`. Every `bin/test-*` script hardcodes that path.
+- Portals: `docker exec -it ivozprovider-portal-<app> bash` (`platform`, `brand`, `client`, `user`); `start.sh` runs `yarn start` in `web/portal/<app>`.
+- Other containers: `ivozprovider-data` (MariaDB), `ivozprovider-redis`.
+- Edit files on the host; the repo is bind-mounted in the backend and portal containers. `../ivoz-ui/` must exist next to the repo.
```
danigargar: This is not exactly a must. You only need it if you want to run the development version of ivoz-ui.

danigargar (changes requested):
[CLAUDE.md](https://github.com/user-attachments/files/31651695/CLAUDE.md)

This CLAUDE.md contains good architectural constraints. Don't discard them. 

I only have doubts about the rule that states "Business logic goes to life-cycle services", because it's a little bit more nuanced. But we can leave it as is until it causes issues, as AI writing a lifecycle for something that usually goes in a sanitizer.

AGENTS.md
```diff
@@ -0,0 +1,148 @@
...
+- `microservices/` — PHP and Go services
+- `asterisk/` — FastAGI app (`agi/`) and Asterisk config
+- `kamailio/` — SIP proxy configs (`users/`, `trunks/`)
+- `cgrates/` — rating engine config
+- `profiles/` — per-host-role filesystem overlays
+- `debian/` — packaging
+- `tests/` — test images, black-box SIP scenarios, API and security tests
+- `doc/` — user docs and `dev/` conventions
+
+## Internal ivoz-* libraries
```
danigargar:
```suggestion
## Shared ivoz-* repositories and libraries
```

"internal" in a monorepo context is ambiguous. It's internal to the company or internal to the monorepo?

AGENTS.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Backend internal packages: `irontec/ivoz-core`, `ivoz-api-bundle` (DDD/API framework: generated entities, DTOs, assemblers, lifecycles), `ivoz-provider-bundle`, `ivoz-dev-tools` (entity generator), `replacements`.
```
danigargar: Resolve this ambiguity:  most of this libraries are internal to the company but ivoz-provider-bundle and replacements that are internal to the monorepo.

danigargar (changes requested):
May be putting whole sections in separate files keeps the context small while telling the agent where to find the information when it needed. 

I don't know if this makes sense. Could you investigate it?

Something like this:
Architecture rules:
See doc/dev/architecture.md

Testing conventions:
See doc/dev/testing.md

Git workflow:
See doc/dev/git.md

AGENTS.md
R0MANDEV: Agreed — "internal" reads as "inside the monorepo", which is the opposite of what these are. Renamed to "Shared ivoz-* repositories and libraries" as suggested. It's in the force-pushed commit rather than applied via the suggestion button, so GitHub will show this thread as outdated.

danigargar (approved):

ironArt3mis (approved):

